PR #38, opened 2026-08-18
Launch antiburn on startup from onboarding

martyportier:
## Summary

- Add a default-on “Launch Antiburn on startup” choice to the final onboarding step.
- Keep the onboarding choice synchronized with General settings.
- Reconcile packaged startup registration on macOS, Windows, and Linux.
- Keep ordinary development builds from changing startup registration.

## Validation

- Frontend and focused onboarding tests passed.
- Vite, TypeScript, ESLint, formatting, Rust tests, and Clippy passed.
- Packaged startup behavior remains scheduled for platform smoke testing.


apps/desktop/src/views/OnboardingView.tsx
```diff
@@ -85,11 +91,31 @@ export function OnboardingView() {
...
+      sawSettingsEvent = true;
+      // A local write emits this event before its invoke promise resolves. An
+      // older event must not overwrite a newer optimistic choice queued behind
+      // it; the final local response below becomes authoritative instead.
+      if (!active || pendingSettingsWrites.current > 0) return;
+      settingsRef.current = stored;
+      confirmedSettingsRef.current = stored;
+      applyTheme(stored.theme);
+      setSettingsState(stored);
+    });
```
daveslutzkin: hm this ref juggling looks a bit messy - wonder if it's actually necessary?

apps/desktop/src/views/OnboardingView.tsx
```diff
@@ -85,11 +91,31 @@ export function OnboardingView() {
...
       const stored = await getSettings().catch(() => DEFAULT_SETTINGS);
       if (!active) return;
-      applyTheme(stored.theme);
-      setSettingsState(stored);
+      if (!sawSettingsEvent && pendingSettingsWrites.current === 0) {
+        settingsRef.current = stored;
+        confirmedSettingsRef.current = stored;
+        applyTheme(stored.theme);
+        setSettingsState(stored);
+      }
```
daveslutzkin: and this

apps/desktop/src/views/OnboardingView.tsx
```diff
@@ -85,11 +91,31 @@ export function OnboardingView() {
   // them; the rest settles together because no step reads another's answer.
   useEffect(() => {
```
daveslutzkin: I need to add guidance to this repo to avoid `useEffect`

daveslutzkin (changes requested):
I'm going to fix the useEffect / settings subscription stuff in my worktree and then push it to this branch

apps/desktop/src/views/OnboardingView.tsx
martyportier: yeah we should.

apps/desktop/src/views/OnboardingView.tsx
martyportier: Resolved in ff9cd9c: onboarding now keeps a local two-field draft and persists once on Finish; the refs and write queue are gone.

apps/desktop/src/views/OnboardingView.tsx
martyportier: Resolved by the same ff9cd9c refactor—the initial event arbitration and its refs are gone.

apps/desktop/src/views/OnboardingView.tsx
martyportier: Agreed. I removed settings synchronization from this effect; what remains is the one-time Tauri window bootstrap.

daveslutzkin (approved):
The onboarding settings/ref concerns are addressed. Onboarding now commits through one atomic finish transition, the view and flow no longer use effects, the repository guidance documents the effect policy, and CI is green.

PR #45, opened 2026-08-18
fix(desktop): live-usage freshness — period markers, visibility-aware refetch, popover-shown refresh

daveslutzkin:
## Weekly and daily elapsed markers from window kind

**What was wrong:** The period-elapsed marker on a usage bar only appeared when the provider stated a window's `startsAt`, or when the window's *id* was in a small hardcoded table (`five-hour` only). A weekly or daily window with a stated reset but no stated start never got a marker at all, even though its period is implied by its own cadence.

**What changed:** `liveWindowElapsed` (in `src/lib/presentation/liveUsage.ts`) now also infers a window's period from its `kind`: `weekly` → 7 days, `daily` → 24 hours, checked after the existing id-keyed table. `monthly` and `billingCycle` are deliberately excluded, since a month's length genuinely varies.

**Why:** A window's cadence is arithmetic on its own identity, the same way `five-hour` already was — `weekly` means seven days by definition, not a guess about where a particular window's boundary falls. Monthly/billing-cycle windows keep no marker because inferring a fixed length there really would be a guess dressed as a measurement.

## Visibility-aware cooldown max-age

**What was wrong:** The popover re-queries `get_live_usage` roughly once a minute while it's open, but every live-usage source held a successful reading behind a fixed 10-minute cooldown (`SUCCESS_COOLDOWN`). A visible, actively-polled panel could therefore sit on a reading up to 10 minutes stale.

**What changed:** `Cooldown::poll` (in `provider_usage/live/sources/cooldown.rs`) now takes a caller-stated `max_age: Duration` in place of the fixed cooldown, threaded through `LiveUsageSource::fetch`, `sources::collect`, and `summarize`. The failure-retry cooldown is derived from the same `max_age`, clamped to **[60s, 300s]**. `get_live_usage` (`commands.rs`) asks for 50 seconds — just under the popover's own polling interval, so an open popover's ordinary polling is what keeps the reading current. The background milestone pass (`usage_alerts.rs`) is unchanged: it still asks for 600 seconds, so background traffic is exactly what it was before.

## Popover-shown event decouples usage refresh from the scan

**What was wrong:** The on-open usage refresh rode the scan pipeline — `note_shown` kicked the scan scheduler, and the frontend refreshed usage on `scan:finished`. Two problems: `ScanController::request()` always wakes the scheduler loop, but the *pass* that wake-up would run is silently skipped whenever discovery is paused or onboarding hasn't finished, which silenced the usage refresh too even though it has nothing to do with local disk discovery; and even when a pass does run, it's a full disk walk the usage refresh had to wait out for no reason.

**What changed:** `note_shown` now also emits `popover:shown` unconditionally, the instant the popover reaches the screen, following the same `module:event` naming and emit mechanism as `scan.rs`'s events. The webview's new `onPopoverShown` listener (`ipc.ts`) refreshes only live usage; entries and the repository list stay on the existing scan-finished path.

---

Rebased onto main after #42 (tauri-nspanel v2.1 upgrade); full verification suite re-run clean on the rebased tree.

### DIFF
--- a/apps/desktop/src-tauri/src/commands.rs
+++ b/apps/desktop/src-tauri/src/commands.rs
@@ -402,6 +402,20 @@ pub fn get_provider_usage(
     Ok(provider_usage::summarize(&evidence, now, offset))
 }
 
+/// How fresh a reading this command asks each source's cooldown for.
+///
+/// This command is called from the popover, which polls it roughly once a
+/// minute for as long as the popover stays visible — see
+/// `scan::TICK` and `popover::note_shown`. Fifty seconds sits just under that
+/// polling interval, so an open popover's own ordinary polling is what keeps
+/// the reading current: every visible tick is close enough to the cooldown's
+/// edge to trigger a real fetch, without this command itself running a timer
+/// or a background task. The aggressive freshness is bounded by someone
+/// actually looking — once the popover closes, nothing here keeps polling on
+/// its behalf, and the background monitor's own, much longer, `max_age`
+/// takes back over (see `usage_alerts::MILESTONE_MAX_AGE`).
+const POPOVER_LIVE_USAGE_MAX_AGE: std::time::Duration = std::time::Duration::from_secs(50);
+
 /// The provider's own limit figures, when a registered source can prove them.
 ///
 /// Deliberately a second command rather than a field on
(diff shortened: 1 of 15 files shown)


martyportier (approved):
Reviewed for high-severity issues, Cadence brand terminology, source-boundary compliance, DCO/provenance, licensing, privacy, and repository governance. Approved.

PR #60, opened 2026-08-19
feat(desktop): consented anonymised usage analytics

martyportier:
## Summary

Add optional anonymized application analytics with a control in onboarding and Settings. The feature reports application health and use, not session content.

- Send nothing before the user completes onboarding and enables analytics.
- Exclude prompts, transcripts, titles, paths, repository names, branches, credentials, token counts, and costs.
- Bucket counts to reduce identification risk.
- Rotate the installation identifier when analytics is disabled and enabled again.
- Keep builds without an analytics endpoint fully inactive.
- Document the event fields and expose the privacy details in the application.

## Validation

- Consent, queue deletion, identifier rotation, schema, and migration tests passed.
- Rust and frontend tests passed.
- Type checks, lint, formatting, and privacy checks passed.


martyportier:
## Verified end to end against a live collector

A distribution build pointed at a local collector, with the real endpoint's contract:

- Unauthenticated `POST /v1/track` accepted — **202**. No auth header needed, which was the open question on this branch.
- `antiburn.app_launched` and `antiburn.scan_completed` queued, flushed, and cleared.
- Payloads carry 10 and 11 leaf fields against the documented ceiling of thirteen (`bucket`, `label`, `detail` are `skip_serializing_if`).
- `scan_completed` reported `bucket: "50-199"` for a real 105-session database — bucketed, not counted.
- No path, repository name, title, token count, or cost in any payload.

Migration `V2 → V3 → V4` verified on that same database.

## One fix, in `4cd7d48`

The run surfaced a startup abort the existing guard cannot catch. `a_database_already_at_v3_still_gets_the_analytics_tables` builds its fixture from `MIGRATIONS[..3]`, which reconstructs main's v3 — the side that won the number. A database carrying the *other* v3 (this branch's analytics tables, stamped 3) runs index 3 against tables already present and fails.

The new test builds that shape and asserts the refusal, because refusing is correct: `IF NOT EXISTS` on V4 would let such a database stamp itself 4 and look fully migrated while having silently skipped main's V3, keeping a stale `liveUsageEnabled` row with nothing left to detect it. A loud failure beats a silent divergence.

The error now names the colliding table — the SQLite cause was always in the source chain, but tauri's setup hook formats with `{}`, so a human saw only `migration 4 failed`.

daveslutzkin (approved):

PR #86, opened 2026-08-20
feat(desktop): add crate-backed floating usage HUD (supersedes #40)

martyportier:
## Summary

Add the floating usage HUD as a standalone desktop crate and integrate it with the current application architecture.

- Preserve compact and expanded HUD states, hover timing, dragging, usage polling, and existing entry points.
- Keep native window code in the HUD crate and application policy in the desktop shell.
- Use an external store for frontend synchronization without React effects.
- Ship the HUD on macOS only. Windows and Linux compile with inactive entry points until their window behavior is implemented and tested.

This PR supersedes PR #40.

## Validation

- Frontend tests, type checks, lint, formatting, and production build passed.
- Rust checks passed on macOS, Linux, and Windows targets.
- Compact and expanded native states were checked manually on macOS.


songcarver:
1. After using the manual X to close the HUD, the switch in the settings is not switched off. 
2. There was a brief flash of a dark semitransparent window appearing on first open. 
3. When turning on and off, it's possible to get the HUD in a state where it no longer responds correctly to hover position. 
4. Text in light mode on the HUD is a little bit too light for legibility. 

None of these are blockers to merge, but I'd like before public release

PR #104, opened 2026-08-21
Enforce the desktop design contract, and fix what enforcing it exposed

martyportier:
## Summary

Add automated checks for the desktop design contract and fix the drift those checks found.

- Verify stylesheet registration, color tokens, theme values, typography, spacing, radii, motion, shadows, window corners, and documented references.
- Add shared motion tokens and a desktop design-review guide.
- Apply the macOS popover material and preserve transparent window surfaces.
- Fix contrast problems across light and dark themes.
- Run the design check for every change that can affect desktop styling.

## Validation

- Design checks and their mutation-tested suite passed.
- Frontend tests, formatting, lint, type checks, build, Rust checks, and accessibility measurements passed.
- The checker detected new stylesheet and palette drift during integration.

### DIFF
--- a/.agents/skills/design-review/design-principles.md
+++ b/.agents/skills/design-review/design-principles.md
@@ -289,10 +289,13 @@ without a comment that says what it means. **[Medium]**
 colour from a CSS variable, so both themes work with no branch in the chart
 code. **[High]**
 
-**RULE 8.3 — a figure is honest about its basis.** Estimated and measured
-figures must be distinguishable, and the label must say which one it is. A
-number the reader can mistake for a provider's own figure is a finding.
-**[High]**
+**RULE 8.3 — a figure is honest about its basis.** A reader must never take an
+estimate for a provider's own figure. Provenance has to be *reachable* — a
+tooltip, a disclosure, a detail row, or one statement covering a whole section
+— but it does not have to be a visible label on every number. On a dense
+surface, per-figure labels usually clutter more than they inform. The finding
+is a figure that reads as authoritative with no way to find out otherwise, not
+a figure without a label. **[High]**
 
 **RULE 8.4 — no chart junk.** One accent, quiet gridlines, a label on the chart
 in place of a busy legend. Thin strokes at chart scale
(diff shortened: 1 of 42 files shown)


apps/desktop/src/styles/tokens.css
```diff
@@ -85,14 +85,19 @@
...
+  --ease-out-quart: cubic-bezier(0.23, 1, 0.32, 1);
 }
 
 /* --- Light (default) --- */
 :root {
   color-scheme: light dark;
 
   /* Backgrounds */
-  --color-bg-primary: rgb(255 255 255 / 0.58);
+  --color-bg-primary: rgb(255 255 255 / 0.85);
```
daveslutzkin, line 100: just an idle wonder, but should we change all colours to hsl from rgb? rgb kinda useless

.agents/skills/design-review/design-principles.md
```diff
@@ -0,0 +1,473 @@
...
+without a comment that says what it means. **[Medium]**
+
+**RULE 8.2 — a chart colour resolves through a token.** Chart series read their
+colour from a CSS variable, so both themes work with no branch in the chart
+code. **[High]**
+
+**RULE 8.3 — a figure is honest about its basis.** Estimated and measured
+figures must be distinguishable, and the label must say which one it is. A
+number the reader can mistake for a provider's own figure is a finding.
+**[High]**
```
daveslutzkin: this is too aggressive and leads to cluttered UIs. there should always be a way to find the provenance but it's not usually a visible label.

daveslutzkin (approved):

.agents/skills/design-review/design-principles.md
martyportier: Agreed — softened in 653f6ed: 8.3 now requires provenance to be *reachable* (tooltip, disclosure, detail row, or one statement per section) rather than a label on every figure, and keeps [High] only for a figure that reads as authoritative with no way to find out otherwise; 12.7 now defers to it.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apps/desktop/src/styles/tokens.css
martyportier: yes we should. i'll put it on my list.

PR #122, opened 2026-08-22
docs: add a token-wastage events reference

songcarver:
Adds `docs/token-wastage-events.md` — a working reference covering eleven events that plausibly waste tokens in a coding-agent session (twelve rows; quota pressure is listed separately, as the ratified catalog keeps it separate).

Each entry records three things:

- **Status today**, grounded against the tree rather than the plan.
- **Whether the event is deterministically detectable** from a transcript, with the actual predicate and its qualifications.
- **What would have to change to avoid it**, split by actor — agent behaviour, codebase, and user configuration.

The doc separates three states that routinely get conflated as "detection": *computed today* (the engine produces the value), *parsed but not judged* (the fact reaches `NormalizedEvent`, nothing decides anything about it), and *not parsed* (the transcript carries the signal, the parser discards it). It also flags throughout where an absence must read as `Unsupported` rather than as zero — nothing in the tree implements `EvidenceValue` yet, so every absence is currently unqualified.

**Verified against `main` at `3a666ce`.** The branch has been merged up to date and every claim re-checked against that revision.

### What the survey found

- **Cache rehydration** is the one finished detector, and the most developed code in this area. It runs **two paths**, selected by `NormalizedSession::cache_write_tokens_available`: the direct rule (`is_cache_rehydration_turn`) for providers that report cache writes, and an inferred rule (`inferred_cache_rehydration_turn`) for providers that do not — that one finds a cache-read collapse followed by a large replay, then waits for the next turn to confirm the cache rebuilt. Seven named thresholds across the two, both parent-only via `EventSource`, both guarding a mid-session model switch via `same_known_model`.
- **Unused MCP servers** and **unused skills** have both halves parsed today — the loaded set from `initial_context.rs` (`mcp_instructions_delta`, `skill_listing`) and the invoked set from `ToolCall::name` / `SkillUse` — and nothing takes the difference between them.
- **Redundant search/re-reading** and **read-back after a confirmed write** are both blocked on one missing field: `ToolCall` keeps only `name`, `category`, and `detail`, so `tool_call_from_input` receives the full input `Value` and discards the target. A stable non-identifying target key would unlock both, and is compatible with the engine's local boundary.
- **Unused built-in tools** is the weakest entry, and the doc says why: `deferred_tools_delta` is flattened into one `SystemInstructions` row, and the always-loaded core tools never appear in the transcript at all.

### Note on staleness

The cache-rehydration entry went stale within two days of first being written — `main` moved and reworked exactly the code the doc anchors to. That is the failure mode a groundedness-claiming document invites, so the file now records the revision its claims were checked against and says plainly that entries need re-verifying when the analysis code moves.

Documentation only — no code, schema, or behaviour changes. The PR diff against its merge base is the single new file.

Written for personal orientation first, so it reads as a working note rather than a specification, and defers to `docs/plans/local-insights-architecture.md` wherever the two disagree.


songcarver:
@codesoda I'm going to close this, but may be interesting to you.

PR #145, opened 2026-08-24
feat(desktop): Show app version and debug status in popover footer

martyportier:
## Summary

- show the running app version beside `antiburn` in the popover footer
- label debug builds with `debug` while release builds show only the version
- keep the version always visible in the secondary text color and omit it if app metadata cannot load
- full time at the moment - in the future we can hide and reveal with modifier again if desired.

## Screenshots
image

## Validation

- `vitest run src/views/PopoverView.test.tsx` — 35 passed
- desktop type-check — passed
- desktop lint — passed
- `cargo test --lib --quiet` — 391 passed
- Rust formatting — passed
- design-contract check — passed
- native Tauri popover inspected under the current system theme

### DIFF
--- a/apps/desktop/src/views/PopoverView.test.tsx
+++ b/apps/desktop/src/views/PopoverView.test.tsx
@@ -205,8 +205,13 @@ function repositoryPayload(overrides: Record<string, unknown> = {}) {
 
 function mockCommands(overrides: Record<string, unknown> = {}) {
   invoke.mockImplementation((command: string, args?: unknown) => {
-    if (command in overrides) return Promise.resolve(overrides[command])
+    if (command in overrides) {
+      const result = overrides[command]
+      return result instanceof Error ? Promise.reject(result) : Promise.resolve(result)
+    }
     switch (command) {
+      case "app_info":
+        return Promise.resolve({ appVersion: "0.1.0", debugBuild: true })
       case "get_settings":
         return Promise.resolve(SETTINGS)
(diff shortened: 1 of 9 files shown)


apps/desktop/src/views/PopoverView.test.tsx
```diff
@@ -451,14 +456,34 @@ describe("PopoverView", () => {
...
 
     const footer = screen.getByRole("button", { name: "Open settings" }).parentElement
     expect(footer).not.toBeNull()
     expect(footer).toHaveTextContent("antiburn")
     expect(footer?.querySelector('[data-testid="usage-limits-bar"]')).toBeNull()
+    expect(screen.getByRole("heading", { name: "antiburn" })).toBeInTheDocument()
+    const version = screen.getByText("v0.1.0 debug")
+    expect(version).toHaveClass("type-caption", "text-label-secondary")
+    expect(version.parentElement).toHaveClass("gap-2")
+  })
```
daveslutzkin, line 471: get rid of the "without putting usage in the footer" bit - that's like saying "without having any elephants". there's no usage there.

daveslutzkin (approved):

apps/desktop/src/views/PopoverView.test.tsx
martyportier: Agreed — removed that clause. The test is now named `shows the app version beside the title`.

Issue #90, opened 2026-08-20
Ratchet the aislop thresholds toward defaults and clear the findings


codesoda:
## Purpose

Lower the `aislop` thresholds toward their defaults, and clear the findings that stand in the way. Companion to #89, which lands the gate with deliberately loose thresholds so it is green on day one.

This is the ratchet. #89 makes the gate real; this issue makes it strict.

## Starting point

Measured on `feat/gh70` at `c03ad1e`, against aislop defaults:

```
57 / 100  Needs Work    0 errors · 33 warnings · 1 fixable
325 files · 5 engines
```

| Count | Rule | Notes |
| --- | --- | --- |
| 19 | `complexity/file-too-large` | The bulk of the gap. Default `maxFileLoc: 400` |
| 15 | `ai-slop/empty-function` | Informational; `ActivityRowPrimitives.tsx:51`, `useActivityGroupPinning.ts:87`, `ipc.ts:1119`, +12 |
| 9 | `complexity/function-too-long` | Default `maxFunctionLoc: 80` |
| 1 | `ai-slop/hardcoded-id` | `provider_usage/live/sources/codex_fetch.rs:87` |
| 1 | `ai-slop/hardcoded-url` | |
| 1 | `ai-slop/meta-comment` | `provider_usage/live/sources/cooldown.rs:306` |
| 1 | `ai-slop/narrative-comment` | Auto-fixable |
| 1 | `ai-slop/rust-non-test-unwrap` | `apps/desktop/src-tauri/crates/sound/src/synth.rs:152` |

Reproduce with `aislop scan`; `aislop scan -d` lists every location.

## Approach: ratchet, not one big pass

Each step is its own PR: lower a threshold in `.aislop/config.yml`, fix what that step newly catches, land it green. Never one enormous refactor, and never a threshold change without the fixes that make it pass.

**Step 0 — free wins, no threshold change.**
- `aislop fix` for the one auto-fixable finding.
- The four single-instance `ai-slop` findings: hardcoded id, hardcoded URL, meta comment, `.unwrap()`. Each is small and self-contained. `synth.rs:152` is the only one with real behavior risk — `AGENTS.md` requires propagating with `?` or an `.expect("context")` naming the invariant, never a silent panic.

**Step 1 — `ai-slop/empty-function` (15, informational).** Each is either an unfinished stub or a deliberate no-op. Deliberate ones want a comment saying why; unfinished ones want implementing. Decide per case; do not blanket-comment to silence the rule.

[...]

Issue #70, opened 2026-08-19
Implement Local Insights from the reviewed architecture plan


codesoda:
## Purpose

Track implementation of Local Insights using the manually researched architecture and sequential implementation guide in #54. This issue is an execution wrapper for SEAMS and does not replace or restate that plan.

## Authoritative planning source

- Plan PR: #54
- Plan file: [`docs/plans/local-insights-architecture.md`](https://github.com/antiburn/antiburn/blob/23979b103624cd8e6dafe9d148e652a64d8053d0/docs/plans/local-insights-architecture.md)
- Pinned revision: `23979b103624cd8e6dafe9d148e652a64d8053d0`

Before implementation begins, use the final reviewed/merged revision of #54 if it differs from the pin above, and record that immutable revision in the SEAMS master plan. Revalidate current-code claims against the implementation base; preserve the plan's researched decisions rather than restarting the architecture research.

## Executable horizon

Implement the first-provider Local Insights release through the plan's Phase 11, using Claude Code as the initial JSONL provider unless the Phase 0 fixture review finds a concrete blocker.

Outcome-level scope:

- Freeze the Claude compatibility baseline and characterization fixtures.
- Add provider-aware source generations and projection revisions.
- Add bounded, recoverable, cancellation-safe JSONL traversal.
- Stream Claude normalization while preserving compatibility.
- Stream existing Claude `SessionMetrics` without observable regression.
- Produce and atomically persist truthful, compact `SessionEvidence` from the same source pass.
- Decouple durable evidence processing from scanning.
- Complete all supported Claude evidence groups and remove every temporary placeholder.
- Build the bounded thirty-day report reducer, nine Hygiene/Efficiency detectors, and separate quota-pressure section.
- Expose truthful coverage, findings, clean, and not-assessed states through desktop IPC and UI.
- Complete privacy, lifecycle, resource, migration, debug, release, and real-session hardening gates.

## Required boundaries

- Persist derived, rule-neutral facts—not raw transcripts, complete canonical sessions/events, or detector conclusions.
- Distinguish `Complete`, `Partial`, and `Unsupported`; incomplete evidence must never produce a false clean result.
- Keep capabilities separate from per-session evidence coverage.
- Reject stale projections when a source changes during processing.
- Update metrics and evidence atomically for the current source generation.
- Keep transcript processing bounded, restart-safe, and outside long-held application database locks.
- Preserve existing session UI behavior until the planned Insights UI slice.
[...]

PR #152, opened 2026-08-24
Stack GH-70: 0003 Source generations and projection revisions

codesoda:
This is part of a PR stack based at #91; this PR merges into main.

### Stack · GH-70
- #91 — GH-70: local-insights architecture (base: master plan)
  - **#NNN — 0003 (CH-002): Source generations and projection revisions ← this PR**

## What this seam contains

Three-slot seam landing the durable source version contract, migration V9, and the scan-side source generation writer:

```mermaid
flowchart TD
    subgraph s0003["Seam 0003 · T3 · Source generations and projection revisions"]
        c1["c1 · Source-version types, the fingerprint composition and its...<br/><i>crates/antiburn-local/src/discovery/source_vers...</i>"]
        c2["c2 · Migration V9, store model, upsert generation arithmetic, ...<br/><i>apps/desktop/src-tauri/src/store/schema.rs, sto...</i>"]
        c3(["c3 · Explorers::source_version producers, descriptor construct...<br/><i>crates/antiburn-local/src/discovery/source_vers...</i><br/>✓ aa1ec2485 (final)"])
        c1 -->|"✓ 1cf58b5ba"| c2
        c2 -->|"✓ 054670eee"| c3
    end
    report["report · approved"]
    s0003 --> report
    classDef done fill:#d4edda,stroke:#28a745,color:#1e4620
    classDef pending fill:#fff3cd,stroke:#b8860b,color:#5c4400
    class c1,c2,c3,report done
```

- **c1**: storage-neutral source-version types (`SourceRead`, fingerprint composition with FNV-1a 64-bit hash on 64 KiB head), and the preview read split into raw head plus lossy conversion.
- **c2**: schema migration V9 (7 new columns on `session` and `session_analysis`), upsert generation arithmetic, store model updates, and migration tests.
- **c3**: `Explorers::source_version` producers for file, provider-DB, and inline sources; canonical `SourceDescriptor` construction; source fingerprint persistence; `scan.rs:describe_one_with_activity` rewired onto the shared `session_log_read` path so it reuses one bounded head read.

## Why this piece was done

CH-002 is the foundation every later Insights scope area reads. No code consumes generations until CH-005/CH-007, making this the cheapest point to land the schema and the source-version contract.

## What it changes

**Storage**: Migration V9 appends 7 columns (`source_fingerprint`, `generation`, `analyzed_generation`, `revision`, `projection_revision_0`, `projection_revision_1`, `projection_revision_2`) to `session` and `session_analysis`. Existing session_analysis rows take `analyzed_generation: 0` and `revision: 1` on upsert. V1–V8 and the analytics cache fingerprint are unchanged.

**Behavior**: Each readable session row now keeps a compact `sv1:` fingerprint of its source and a generation counter that increases only when that fingerprint changes. A read failure, stat failure, or absent provider fingerprint yields a `NULL` fingerprint; the upsert preserves the stored fingerprint and stored generation, so an unreadable source is never a new generation.

Content is now built only for consumers that use it: ProviderDb renders and Inline clones only for Claude and Codex consumers. Antigravity, Cursor, and OpenCode sessions no longer pay a discarded full-transcript cost per scan. File-backed sources keep the one-read contract: the fingerprint uses the bounded head read that describing a session already performs, and for Claude/Codex sessions the second read of that same head is removed.

**Divergences from plan** (four behavior-preserving, one user-visible fix):
1. The approved plan (line 175) explicitly specified that Inline and ProviderDb keep `SourceRead.content` unset. Following that plan caused a user-visible regression in Claude Desktop Inline sessions: subagent detection, title fallback, fork-parent extraction all depend on preview content. The fix deliberately diverges: content is now set for Claude/Codex consumers only, restoring the old behavior for those agents while keeping non-consuming agents free. This is a **behavior-preserving fix for a regression in the approved plan's specification**.
2. `scan.rs:top_up_analysis` signature changed to inject `session_log_read` result instead of taking a path. Behavior-preserving, test-enabling.
3. Inline and ProviderDb preview content building was narrowed to Claude/Codex. Behavior-preserving.
4. Inline clone guard added (same consumer cohort as ProviderDb render). Behavior-preserving.

## How to review

Read in order:
1. `crates/antiburn-local/src/discovery/source_version.rs` — new types (`SourceRead`, `SourceDescriptor`, `Explorers::source_version` producers for each source kind).
2. `crates/antiburn-local/src/discovery/mod.rs` — integration into `session_source_metadata` and `session_source_content`, fingerprint/descriptor building, new tests.
3. `apps/desktop/src-tauri/src/store/schema.rs` — migration V9 and column definitions.
4. `apps/desktop/src-tauri/src/store/model.rs` — `SessionAnalysis` structure and generation fields.
5. `apps/desktop/src-tauri/src/store/mod.rs` — `session_from_row` and `upsert_session_in` generation arithmetic and fingerprint upsert.
6. `apps/desktop/src-tauri/src/scan.rs` — `describe_one_with_activity` rewrite onto `session_log_read` path, injection of log-read result into `top_up_analysis`, content guard for Claude/Codex consumers.
7. Tests in each crate.

Verify: The one-read contract for file sources (fingerprint reuses the head read from describe-a-session). The generation-preserving upsert for unreadable sources. Claude/Codex content building; non-consuming agents discard no content. Windows `cfg(windows)` fixture integration at line 483 (cannot compile/run on macOS; Windows CI is its first execution).

**Review hotspots** — risk triage of source entities in this diff. Hand-written, source-only.

| Entity | File | Risk | Signals |
|---|---|---|---|
| `SourceRead` (struct, new) | `crates/antiburn-local/src/discovery/source_version.rs` | Medium (0.48) | new public type · discoverer contract boundary · fingerprint/content fields used by scan and store |
| `SourceDescriptor` (struct, new) | `crates/antiburn-local/src/discovery/source_version.rs` | Medium (0.44) | new persisted descriptor · canonical source identity across session lifetime |
| `Explorers::source_version` (fn, new) | `crates/antiburn-local/src/discovery/mod.rs` | Medium (0.46) | new producer · reads every session source three ways · blast 2 (Inline, file, ProviderDb arms call distinct content paths) |
| `describe_one_with_activity` (fn, modified) | `apps/desktop/src-tauri/src/scan.rs` | High (0.62) | scan entry point rewrite · behavior-critical · 437 line diff · reuses head read on session_log_read path · content consumer guard for Claude/Codex · regression fix from approved plan |
| `upsert_session_in` (fn, modified) | `apps/desktop/src-tauri/src/store/mod.rs` | High (0.58) | generation upsert arithmetic · persisted generation tracking · unreadable-source semantics (preserves old generation on NULL fingerprint) |
| `session_from_row` (fn, modified) | `apps/desktop/src-tauri/src/store/mod.rs` | Medium (0.42) | reads new fingerprint/generation columns from session row · used on every session load |
| Migration V9 (schema.rs) | `apps/desktop/src-tauri/src/store/schema.rs` | High (0.64) | schema migration on live databases · 7 new columns · persisted data shape change · deployed to every session row |

## Verification

- Crates/antiburn-local: 694 tests, fmt clean, clippy -D warnings clean.
- Apps/desktop/src-tauri: 400 tests, fmt clean, clippy -D warnings clean.
- `pnpm run slop`: exits 1 on five standing issue #90 findings (none in this seam's diff) plus two pre-existing non-blocking warnings in an untouched file. Per AGENTS.md this seam must not fix standing #90 findings.
- **Windows CI risk**: the `cfg(windows)` SourceStat fixture `a_source_stat_reports_identity_and_change_time` (crates/antiburn-local/src/discovery/source_version.rs line 483) cannot compile on the macOS dev checkout (E0463: missing x86_64-pc-windows-msvc std target). **Windows CI is its first real execution.** Reviewers should watch the Windows jobs specifically for this fixture.

### DIFF
--- a/apps/desktop/src-tauri/src/scan.rs
+++ b/apps/desktop/src-tauri/src/scan.rs
@@ -632,6 +632,7 @@ async fn describe_one_with_activity(
         activity_source,
         subagent_count,
         fork_parent_session_id,
+        source_fingerprint: None,
     }))
 }
 
@@ -1741,6 +1742,7 @@ mod tests {
             activity_source: "mtime".into(),
             subagent_count: 0,
             fork_parent_session_id: None,
+            source_fingerprint: None,
         }
     }
 
--- a/apps/desktop/src-tauri/src/scan.rs
+++ b/apps/desktop/src-tauri/src/scan.rs
@@ -71,8 +71,8 @@ use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use antiburn_local::discovery::scanner::{self, TitleSource};
 use antiburn_local::discovery::{
-    Explorers, ResolvedTitle, SessionLog, SessionSource, TitleLookupKind, session_log_metadata,
-    session_source_preview, session_source_tail,
+    Explorers, ResolvedTitle, SessionLog, SessionSource, SourceDescriptor, TitleLookupKind,
+    session_log_read, session_source_tail,
 };
 use antiburn_local::model::AgentKind;
 use antiburn_local::paths::{home_dir, ignored_paths};
@@ -326,7 +326,8 @@ async fn pass(app: &AppHandle, activity_window_days: Option<u32>) -> anyhow::Res
 
     // Everything discovered so far is already persisted, so a cancel here keeps
     // the reader's results and only skips the work still ahead.
-    if app.state::<ScanController>().cancelled() {
+    let controller = app.state::<ScanController>();
+    if controller.cancelled() {
         return Ok(records.len());
     }
 
@@ -338,9 +339,21 @@ async fn pass(app: &AppHandle, activity_window_days: Option<u32>) -> anyhow::Res
             crate::store::MIN_ACTIVITY_DAYS,
             crate::store::MAX_ACTIVITY_DAYS,
         );
-    top_up_analysis(app, now, i64::from(activity_window_days)).await?;
+    top_up_analysis(
+        &store,
+        &controller,
+        now,
+        i64::from(activity_window_days),
+        |agent, session_id, wsl_distro| async move {
+            analytics::locate(agent, &session_id, wsl_distro.as_deref()).await
+        },
+        |agent, session_id, wsl_distro| async move {
+            analytics::analyze(agent, &session_id, wsl_distro.as_deref()).await
+        },
+    )
+    .await?;
 
-    if app.state::<ScanController>().cancelled() {
+    if controller.cancelled() {
         return Ok(records.len());
     }
 
@@ -556,9 +569,9 @@ async fn describe_one_with_activity(
     indexed_title: Option<ResolvedTitle>,
     activity_state: Option<SessionActivityState>,
 ) -> DescribeOutcome {
-    let metadata = session_log_metadata(&log).await;
+    let read = session_log_read(&log).await;
+    let metadata = read.as_ref().map(|read| &read.metadata);
     let Some(session_id) = metadata
-        .as_ref()
         .and_then(|metadata| metadata.session_id.clone())
         .or_else(|| recovered_id(&log))
     else {
@@ -573,12 +586,13 @@ async fn describe_one_with_activity(
         log.agent_type.slug(),
         session_id.clone(),
     );
-    // One bounded content read serves both content checks below.
     let preview = match log.agent_type {
-        AgentKind::Claude | AgentKind::Codex => session_source_preview(&log.source).await,
+        AgentKind::Claude | AgentKind::Codex => {
+            read.as_ref().and_then(|read| read.content.as_deref())
+        }
         _ => None,
     };
-    if is_subagent_transcript(&log, preview.as_deref()) {
+    if is_subagent_transcript(&log, preview) {
         return DescribeOutcome::Subagent(key);
     }
 
@@ -589,12 +603,10 @@ async fn describe_one_with_activity(
     };
     let (title, title_source) = select_title_pair(
         resolved_title,
-        metadata
-            .as_ref()
-            .and_then(|metadata| metadata.title.clone()),
-        metadata.as_ref().and_then(|metadata| metadata.title_source),
+        metadata.and_then(|metadata| metadata.title.clone()),
+        metadata.and_then(|metadata| metadata.title_source),
         &log.agent_type,
-        preview.as_deref(),
+        preview,
     );
 
     // A dir listing per orchestrator-capable session; vendors that record no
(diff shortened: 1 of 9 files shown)


apps/desktop/src-tauri/src/store/tests.rs
```diff
@@ -930,6 +938,194 @@ fn migrating_forward_drops_a_legacy_live_usage_off_row_so_the_new_default_applie
...
+                   WHERE name IN ('analyzed_generation', 'parser_revision',
+                                  'analyzer_revision', 'metrics_schema_revision')",
+                [],
+                |row| row.get(0),
+            )
+            .unwrap();
+        assert_eq!(added_columns, 3, "start version {start}");
+        assert_eq!(projection_columns, 4, "start version {start}");
+    }
+}
```
daveslutzkin, line 982: is this test necessary? seems heavyweight

apps/desktop/src-tauri/src/store/tests.rs
```diff
@@ -930,6 +938,194 @@ fn migrating_forward_drops_a_legacy_live_usage_off_row_so_the_new_default_applie
...
+                    metrics_schema_revision
+               FROM session_analysis
+              WHERE session_id = 'legacy'",
+            [],
+            |row| Ok((row.get(0)?, row.get(1)?, row.get(2)?, row.get(3)?)),
+        )
+        .unwrap();
+
+    assert_eq!(revisions, (0, 1, 1, 1));
+}
```
daveslutzkin, line 1021: this isn't necessary, this is pre-production so there's basically no existing rows that need migrating. this test will hang around forever.

apps/desktop/src-tauri/src/store/tests.rs
```diff
@@ -930,6 +938,194 @@ fn migrating_forward_drops_a_legacy_live_usage_off_row_so_the_new_default_applie
...
+    store.upsert_sessions(&[rescanned]).unwrap();
+
+    assert_eq!(store.session_count().unwrap(), 1);
+    let after = store
+        .session_source_state(&key)
+        .unwrap()
+        .expect("source state");
+    assert_eq!(after.source_generation, 1);
+    assert_eq!(after.source_fingerprint.as_deref(), Some("sv1:stable"));
+}
```
daveslutzkin, line 1070: get rid of this test too

apps/desktop/src-tauri/src/scan.rs
```diff
@@ -1267,6 +1344,307 @@ mod tests {
...
+            },
+            updated_at: None,
+            environment: DiscoveryEnvironment::Native,
+        };
+
+        assert!(matches!(
+            describe_one(log, std::path::Path::new("/tmp"), None).await,
+            DescribeOutcome::Skip
+        ));
+    }
```
daveslutzkin, line 1492: this should be called `an_empty_session_id_is_skipped` rather than `an_empty_session_id_is_still_skipped` - get rid of the "still" because that just refers to a past state which will be irrelevant in the future

daveslutzkin (approved):

apps/desktop/src-tauri/src/store/tests.rs
codesoda: Done for the other three: both migration tests are gone and the `still` is dropped from the rename.

Kept this one, though happy to remove it if you'd still rather. It reads like a migration test next to the others, but it isn't: it upserts the same record three times (same fingerprint twice, then changed, then None) and asserts `source_generation` only increments when the fingerprint changes. That's the generation arithmetic this seam adds, not legacy-row handling, so the pre-production argument doesn't quite cover it. Dropping it leaves that contract without a direct guard. Say the word and it goes.

PR #167, opened 2026-08-24
Name Codex sessions with Apple's on-device model

songcarver:
Codex sessions whose only name is their raw first user message now get a real title on macOS, generated by Apple's on-device Foundation Models. No API tokens are spent; nothing leaves the machine.

Screenshot 2026-08-25 at 09-02-44

## What changed

**Step 1 — cleaner fallback everywhere.** `clean_first_message_title` strips `[Image #1]`-style attachment markers, keeps the first sentence, and truncates at a word boundary. All platforms benefit even with the model off.

**Step 2 — `TitleSummarizer` trait and `localSummary` provenance.** A new title source sits between `firstMessage` and real vendor titles: it survives first-message re-upserts but loses to a user rename or a vendor title. Store writes are guarded so a better title arriving mid-pass wins. The scan collects candidates (native Codex sessions stuck on `firstMessage`) with up to 3 later user messages as context.

**Step 3 — the macOS backend.** A small Swift sidecar (`sidecar/title-summarizer.swift`) talks to `FoundationModels`: `--probe` answers availability; a plain run reads the session's opening messages as JSON on stdin and prints a 3–6-word title. `build.rs` compiles it for Apple targets and `tauri.macos.conf.json` bundles it beside the app executable. Generation is one process run per title with a 20s timeout, ~300ms–1s each in practice.

## Guardrails

- **Generate once per session** — a session already on `localSummary` is never re-titled, so the 60s scan tick does no repeat work.
- **15 titles per pass, newest first** — a first scan with hundreds of fallback sessions catches up over a few passes instead of stalling one.
- **Silent failure** — model unavailable, timeout, or a refusal-shaped reply all leave the cleaned fallback title in place; a later pass retries.
- **Setting** (`localSummaryTitles`, default on) with a macOS-only toggle in General settings.

## Out of scope (per the approved plan)

Windows Phi Silica, an Ollama backend, shipping model weights, and titling non-Codex agents. The plan's regenerate-on-growth idea was simplified to generate-once. Plan: `docs/plans/local-session-titles.md`.


songcarver:
Ran the shipped pipeline against the real Codex store on this machine (329 threads) to produce before/after examples — and it caught a bug: Codex records `# AGENTS.md instructions …` and `<environment_context>` as **user-role turns**, so the candidate collector fed the model the boilerplate instead of the first real prompt. Every session in a repo with an AGENTS.md produced titles like "Complete AGENTS.md" or "Cadence CLI Setup".

Fixed in fb02a29: candidate collection now skips injected turns (prefix check: `<` or `# AGENTS.md instructions`), with tests.

Real examples after the fix — raw first message vs. cleaned fallback (all platforms) vs. Apple on-device title (macOS), generated with the actual compiled sidecar:

| Raw first message (before) | Cleaned fallback | On-device title |
|---|---|---|
| `pull latest main. start a new branch\nUx issue: Backfill command should require being logged in` | pull latest main | Create new branch for Ux issue |
| `Continueing work on tasks/handoff-2026-03-12.md and README.md\n\nRead those\n\nThen, adjust view so onlyone 1 alert…` | Continueing work on tasks/handoff-2026-03-12.md and… | Continue work on files |
| `ERR_PNPM_NO_IMPORTER_MANIFEST_FOUND No package.json (or package.yaml…) was found in "/Users/keithlang"…` | ERR_PNPM_NO_IMPORTER_MANIFEST_FOUND No package.json (or… | Investigate dependency issue |
| `IMPORTANT: Active note path: …/Customer Notes Synthesis: Travis + Matt + Team.md When the user refers to "this file"…` | IMPORTANT: Active note path:… | Resolve customer notes synthesis |
| `create an 'itunes' like music visualizer, hipnotic  trippy visuals` | create an 'itunes' like music visualizer, hipnotic trippy… | Create music visualizer |
| `[Image #1] in this pane, it should be possible to click on a session to expand it to show more detail` | in this pane, it should be possible to click on a session… | Expand session detail |

z0w0 (changes requested):
Considering this is a local app, and we're going to the effort to bundle in a sidecar executable: I think it would be much better if the executable was a more generic "run-foundation-model" or something similar that could take in a prompt and then output of some kind. Then we could use it in other places easily. It feels overkill to have an executable just for summarizing one specific field

songcarver:
Done in c0bbec6 — the sidecar is now `run-foundation-model`, a generic runner: stdin `{"instructions", "prompt"}` JSON → stdout model response, plus the `--probe` availability check. The title-specific prompt moved into Rust (`antiburn_local::titles::TITLE_INSTRUCTIONS` / `title_prompt`), so other features can reuse the executable as-is.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

martyportier:
Set back to draft as discussed @z0w0 and @songcarver as i'll finalise a potential deterministic approach as the default direction. 

this still stays as a viable option - and certainly if we want more enriched titles down the track. However, the following should be fixed first (validate this is in fact an issue):

1. High — title generation can block scanning for minutes. The scan directly awaits the title pass, which runs up to 15 model processes sequentially with a 20-second timeout each. A degraded model could delay session updates by roughly five minutes, then retry failures on the next scan. Move this work to a background queue with a failure cooldown. ([scan](https://github.com/antiburn/antiburn/blob/5a591abee4ea7370f9af7ed833b569a9458e24f8/apps/desktop/src-tauri/src/scan.rs#L359-L375), [title loop](https://github.com/antiburn/antiburn/blob/5a591abee4ea7370f9af7ed833b569a9458e24f8/apps/desktop/src-tauri/src/titles.rs#L147-L198))

songcarver:
Validated and fixed the blocking concern in 90e2988. Confirmed real: the scan awaited `local_summary_pass` directly, so 15 candidates × 20 s timeout could hold the rest of the pass ~5 minutes on a degraded model.

Now the scan just sorts the batch (newest first) and enqueues it on a dedicated title worker (`TitleWorkerHandle`, spawned alongside the insights worker) — model time never blocks scanning. A pass that spends model/probe time and writes nothing starts a 10-minute failure cooldown, so a degraded model is retried slowly instead of on every scan. Batches are replace-on-enqueue; nothing is lost since the next scan re-collects any session still on the fallback.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #207, opened 2026-08-26
feat(desktop): add the hotspot block to the activity popover

songcarver:
## What

Adds the hotspot block at the foot of the activity popover. It names the most
common Hygiene & Efficiency finding across the assessed 30-day cohort, says how
many sessions carry it, hands over one pasteable fix, and opens to show the
evidence behind the count.

Presentation only. No detector can produce a finding yet, so `PopoverView`
passes `null` and **the block renders nothing on this branch**. The contract in
`hotspot.ts` is frozen first because the presentation is the part under design
review, and freezing the shape lets the two halves land in either order.

## Screenshots

Both shots are the real component at the popover's 380px, rendered against a
temporary sample finding that is **not** in the commit.

Closed the block is 68px. Open with five evidence rows it is 236px, taken out
of the session list, which already scrolls. Past `max-h-56` (about eight rows)
the detail scrolls in place, so the block never grows further and the window
never resizes.

Dark mode was checked in the running app at 400x700 and is not pictured. The
block adds no colour of its own, so both themes come from tokens the contract
already carries.

## Three rules the block keeps

- **No finding, no block.** With nothing to say it renders `null` and the
  session list takes the space back. Not an empty shell and not an "all clear"
  message: a cohort that is only half read must never render as clean (FR-14),
  and nothing at all cannot be misread.
- **The window never resizes.** The activity surface stays at
  `DEFAULT_POPOVER_HEIGHT`. A test pins `POPOVER_HEIGHTS.activity` so a later
  change cannot quietly make the block ask the shell for a taller window and
  move the tray popover under the cursor.
- **One orange mark, and it is the rule across the top.** The fix field is a
  plain stroked field and the count is ordinary label text. An earlier build
  filled the field with the brand colour and set the count in `text-brand`;
  both were too loud. No new token, and `check-design-drift.mjs` passes
  untouched.

## Notes for review

- **No `useEffect`.** The copy check reverts on a `setTimeout` the click starts,
  which is work the click owns. A timer that outlives the block sets state on an
  unmounted component, which React ignores.
- **A rejected clipboard write is caught and shows nothing.** A webview with no
  user activation rejects `writeText`; this was hit for real in `pnpm dev:web`.
  The reader can still select the line, and a 26px field has no room to say
  more. The test for it asserts only the visible half — that the rejection is
  *handled* is enforced by vitest, which reports an unhandled rejection and
  exits non-zero. Drop the `.catch` and the file fails on that route.
- **The whole fix field is the copy target**, not the 13px icon beside it.

## Checks

Green on `main` at `684b0a4`: `type-check`, `lint`, `format`, `knip`,
`check-design-drift.mjs`, `pnpm run slop` (0 errors, 0 warnings), and 849 unit
tests in 84 files. `HotspotBlock.test.tsx` adds 12 of those.

## Follow-ups, not in this PR

- The nine detectors and the evidence groups they need (GH-70).
- `get_local_insights_report`, the winner selection, and swapping the `null`.

Plan: `docs/plans/hotspot-popover-block.md`

>
Screenshot 2026-08-26 at 14-18-00

### DIFF
--- a/apps/desktop/src/views/PopoverView.tsx
+++ b/apps/desktop/src/views/PopoverView.tsx
@@ -22,6 +22,7 @@ import {
   type ProviderUsageSummaryPayload,
 } from "../lib/ipc"
 import type { PopoverSurface } from "../lib/popoverHeight"
+import { HotspotBlock } from "./popover/HotspotBlock"
 import { PopoverSession, sessionKey } from "./popover/PopoverSession"
 import { UsageView } from "./popover/UsageView"
 import type { SessionSubject } from "./popover/SessionPane"
@@ -369,6 +370,13 @@ export function PopoverView() {
             />
           )}
         </div>
+
+        {/* The hotspot block sits between the list and the footer, and shows
+            nothing until the local insights report exists
+            (docs/plans/hotspot-popover-block.md). The list above it scrolls,
+            so a block that opens takes rows off the list and the window keeps
+            its height. */}
+        <HotspotBlock finding={null} />
 
         <PopoverFooter
           appVersion={state.appVersion}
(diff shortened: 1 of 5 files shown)


z0w0:
Closing before the repository history reset. The work remains preserved in the private local archive.

Issue #222, opened 2026-08-27
Recognize newer Claude Code housekeeping record types


codesoda:
# T2 — Recognize newer Claude Code housekeeping record types

**Kind:** fix · **Shape:** directly implementable — the research is done (real-cohort diagnostics, 2026-08-27)
**Origin:** `docs/plans/local-insights-followups.md` — "Additional Claude JSONL row types" (GH-70 seam 0001), upgraded by a field report from a 65-session cohort.

## Problem

`is_recognized_eventless` (`crates/antiburn-local/src/analysis/vendors/jsonl.rs:272`) accepts only `attachment`, `summary`, `file-history-snapshot`. Newer Claude Code writes ~13 more housekeeping record types. Any session containing one degrades to `Partial(UnrecognizedRecordType)`, and under FR-14 partial evidence can never support "clean" — so **five detectors report "evidence is incomplete" for every session of every current Claude Code user**. The detectors find nothing wrong; they just refuse to say so.

Observed frequency across the 65-session cohort:

| type | sessions | type | sessions |
|---|---|---|---|
| `permission-mode` | 65 | `pr-link` | 29 |
| `mode` | 65 | `atis-latch` | 29 |
| `last-prompt` | 64 | `worktree-state` | 9 |
| `ai-title` | 63 | `relocated` | 9 |
| `queue-operation` | 53 | `frame-link` | 7 |
| `file-history-delta` | 33 | `cost-state` | 4 |
| | | `agent-name` | 1 |

## Scope

1. Add the 13 types to the recognized-eventless allowlist. Before adding each, confirm against the harness knowledge base (`crates/harness-kb/facts/claude.json`) that it carries no usage data — the field report says none do; verify, don't inherit.
2. One synthetic hand-authored fixture line per type in `crates/antiburn-local/tests/fixtures/claude_characterization/`, plus rows in the README's "File | Proves" table (that table is already missing three existing fixtures — close that too).
3. Bump `parser_revision` — revision reconciliation requeues existing sessions; on reparse they flip to `Complete` and the five detectors start reporting Clean/Findings.
4. Acceptance: a fixture cohort containing all 13 types produces `Complete` coverage and real detector statuses.

## Decision to record (policy, not code)

[...]

PR #239, opened 2026-08-27
Measure the Claude insights pipeline and record a baseline (#224)

### LINKED ISSUE #224: Full-pipeline measurement & profiling, then act on the numbers
# T5 — Full-pipeline measurement & profiling, then act on the numbers

**Kind:** carved out of CH-013 (the measurement component of the release-hardening item)
**Origin:** master plan `docs/plans/local-insights-architecture.md` CH-013 measurement clause, plus three followups-ledger entries this ticket absorbs as named follow-ups: "Phase 13 optimizations", "Evidence for the Claude append-only guarantee" (seams 0007/0008), "Fork-job transcript materialization during discovery" (seam 0008), and the folded entry "Slow discovery while the popover is hidden".

## Part 1 — full-process integration tests

Integration-type tests that exercise the whole pipeline end to end over synthetic sources: **transcript parsing → normalization → metrics + evidence accumulation → dual-projection persistence to the database** (and on through report query → reduction where the scenario warrants). Scenario matrix:

- a representative many-record session
- a single-oversized-line source (8 MiB retained-record bound exercised)
- a large source (≫ retained buffer bound)
- a parent + subagent roster
- an **actively-written source** (writer appending during the read — the `SourceChanged` full-reprocess path, since Claude's append-only guarantee is Absent)
- a provider-DB-backed source

These double as regression armor for the seams' individual guarantees, now proven in composition.

## Part 2 — `cargo bench` memory/performance profiling

Benches over the same scenarios, measuring **separately** (the ten stages CH-013 named):

1. discovery · 2. queue wait (pending→claim) · 3. source reading · 4. parsing · 5. metrics accumulation · 6. evidence accumulation · 7. persistence · 8. report query · 9. reduction · 10. IPC

Plus the four memory/behavior figures:

- peak retained line buffer (must respect the 8 MiB bound)
- accumulator memory (per-session retained state — put a number on CH-005's documented "proportional growth")
- peak report memory (30-day reduction accumulator)
- **active-writer impact**: `SourceChanged` rejection/retry rate while a representative session is being written

## Part 3 — act on the numbers

- **Tune or explicitly reaffirm** worker concurrency (1 CPU / 1 source / 1 provider-DB permits) and lease values (`LEASE_SECS = 300`, `LEASE_RENEW_SECS = 60`, `insights_worker.rs`). The plan's bias stands: a background utility competing with the reader's live agent sessions is worse than a slow backlog — but "keep 1/1/1" must be a measured outcome, not an inherited default.
- **Named follow-up candidates, each resolved per its number:**
  1. *Phase-13 optimizations* (report caching, relational evidence projections, read pooling) — justified only if a stage timing shows a bottleneck; otherwise dismissed with the number as the reason.
  2. *Claude append-only guarantee evidence* — pursue only if the active-writer rejection rate is materially non-zero. If near zero, the followups entry's own words apply: "the work may never be justified." (Flipping Claude to `Evidenced` is a one-function change; the evidence itself must come from real pinned CLI versions via the harness KB, outside this repo.)
  3. *Fork-job `Inline` materialization* — fork-job discovery reads whole transcripts into memory upstream of the streaming boundary (`discovery/agents/claude.rs`); act per the measured affected-source volume.
  4. *Popover-hidden staleness review* — measure how stale metrics/evidence get while scheduled scanning is paused, and review whether that changes anything (a reversal would require Locked Decision 15 to change — review, not implement).

## Constraints

- Synthetic sources only; benches must not read real transcripts.
- Bench artifacts and numbers land in the ticket/PR, not in evidence tables.

codesoda:
Closes #224

## Summary

Measures the Claude insights pipeline end to end and records a baseline, using only synthetic, seeded, fictional content:

- **Synthetic corpus generator** (`tests/support/corpus.rs`) — deterministic, seeded generator for JSONL transcripts and provider-DB (SQLite) sources; nothing reads a real transcript.
- **Integration tests** — `tests/pipeline_corpus.rs` runs the full in-crate pipeline (source read → framing → parse/normalize → metrics + evidence accumulation → report reduction) over the corpus, including the provider-DB-backed `RawSource::Sqlite` path; `tests/claude_characterization.rs` pins vendor-format behavior against goldens.
- **Criterion bench targets** (`benches/pipeline_baseline.rs`) — timing and memory measurement of every in-crate stage, with results and analysis recorded in `benches/BASELINE.md`. Numbers are indicative local baselines, not CI-enforced thresholds.

Stages the master plan names that live in the desktop app (discovery, queue wait, persistence, report query, IPC) need a desktop-side harness and are the follow-up half of the measurement work.

## Headline baseline numbers

Machine context: Apple M3 Pro (11 cores, arm64), 18 GiB RAM, macOS 26.2, rustc 1.97.0, optimized bench profile, 2026-02 run.

| Measurement | Result |
|---|---|
| Framing throughput (`BoundedJsonlReader`, 6 MiB of small lines) | 3.05 ms (~2.07 GiB/s) |
| Full reparse, 1 MiB session | 7.38 ms (~146 MiB/s) |
| Full reparse, 10 MiB session | 72.5 ms (~149 MiB/s) |
| Full reparse, 50 MiB session | 351 ms (~154 MiB/s) — linear at ~150 MiB/s |
| Per-stage split at 10 MiB | serde parsing dominates (~98 %); metrics + evidence accumulators add only ~2–3 % |
| Provider-DB (SQLite) walk, 20k rows / ~6.2 MiB | 30.7 ms (~205 MiB/s) — linear, slightly faster per byte than JSONL |
| Report reduction | ~0.75 µs/session (65-session field cohort: 49.2 µs; 500 sessions: 367 µs) |
| Framing memory high-water (10 MiB of small lines) | 446 bytes; single near-8 MiB line stays under the `MAX_RECORD_BYTES` bound |
| Metrics accumulator retention, 10 MiB dense source | 10.4 MB (~303 bytes/turn, ≈ 1.0× source bytes) |
| Active-writer `SourceChanged` rejections | 15/15 at 2 ms append interval; 0/15 at ≥ 20 ms; rate ≈ read_window / write_interval, effectively zero for realistic write spacing |

## Recommendations on the four measurement questions

1. **Phase-13 optimizations (report caching, relational evidence projections, read pooling)** — *dismiss with the number.* Report reduction is ~0.75 µs/session; no stage shows a bottleneck those changes would relieve. The only material cost is serde parsing at ~150 MiB/s, and even a 50 MiB session completes in 0.35 s.
2. **Claude append-only guarantee evidence** — *defer with evidence; likely never justified.* Full reprocess is linear and cheap (7 ms / 73 ms / 351 ms at 1 / 10 / 50 MiB), and the `SourceChanged` rejection rate reached zero in every scenario with realistic write spacing. Incremental byte-offset parsing would save at most a fraction of ~0.35 s per pathological session.
3. **Fork-job `Inline` materialization** — *defer.* Materializing costs ~6 % extra time plus one transient full-source buffer per fork job; with the single source permit, peak transient memory equals the largest single session. Revisit only if field sessions well beyond 50 MiB appear.
4. **Popover-hidden staleness (Locked Decision 15 review)** — *reaffirm; no change proposed.* Catch-up drain once the popover opens is N × per-session cost: the 65-session field cohort drains in well under one second on one permit; even a pathological 65 × 10 MiB backlog drains in ~5 s.

Worker concurrency and leases (Part 3 tuning clause): keep 1 CPU / 1 source / 1 provider-DB permit and `LEASE_SECS = 300` / `LEASE_RENEW_SECS = 60` — now a measured outcome rather than a guess.

## Verification

| Gate | Command | Result |
|---|---|---|
| engine: cargo fmt --check | `cd crates/antiburn-local && cargo fmt --check` | ✅ No formatting diffs |
| engine: cargo clippy | `cargo clippy --all-targets --locked -- -D warnings` | ✅ Clean; no warnings |
| engine: cargo nextest | `cargo nextest run --locked --lib --tests` | ✅ 929 passed, 1 skipped; suite wall time 3.49s — new integration tests do not blow up duration |
| engine: bench compile | `cargo bench --no-run` | ✅ Both bench executables built (lib benches + `benches/pipeline_baseline.rs`) |
| desktop | applicability check via `git diff --name-only origin/main..HEAD` | N/A — no `apps/desktop` files in the diff; desktop CI jobs would not trigger |
| repo: CI control-plane tests | `node --test scripts/*.test.mjs` (6 suites) | ✅ 76/76 pass after `pnpm install --frozen-lockfile` |
| repo: design-contract drift | `node scripts/check-design-drift.mjs` | ✅ Design contract in sync with the stylesheets |

## Merge-order note

This branch avoids every file the open PRs #231, #232, and #234 touch, so it merges cleanly in any order relative to them.

### DIFF
--- a/crates/antiburn-local/Cargo.lock
+++ b/crates/antiburn-local/Cargo.lock
@@ -15,11 +15,33 @@ dependencies = [
 ]
 
 [[package]]
+name = "aho-corasick"
+version = "1.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c982642fa9e8606056828ee9a8505737230110bb1099153c79efe865c59d12ba"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
+
+[[package]]
+name = "anstyle"
+version = "1.0.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "940b3a0ca603d1eade50a4846a2afffd5ef57a9feac2c0e2ec2e14f9ead76000"
+
+[[package]]
 name = "antiburn-local"
 version = "0.1.9"
 dependencies = [
  "anyhow",
  "async-trait",
+ "criterion",
  "filetime",
  "futures-util",
  "libc",
@@ -53,6 +75,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "autocfg"
+version = "1.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f2032f911046de80f0a198e0901378627c33f59ea0ac00e363d481118bd70a53"
+
+[[package]]
 name = "bitflags"
 version = "2.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -71,6 +99,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc652a48c352aef3ea3aed32080501cf3ef6ed5da78602a020c991775b0aff04"
 
 [[package]]
+name = "cast"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
+
+[[package]]
 name = "cc"
 version = "1.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


codesoda:
Added a peak-heap comparison answering "how does streaming JSONL compare to reading whole files into memory" (`benches/memory_baseline.rs`, separate binary so the counting allocator does not perturb the timing baseline):

| Source | Streaming peak | Inline peak | Inline / streaming |
|---|---|---|---|
| 1 MiB | 1.66 MB (1.49× source) | 2.68 MB (2.40×) | 1.61× |
| 10 MiB | 20.7 MB (1.84× source) | 32.0 MB (2.84×) | 1.54× |
| 50 MiB | 98.9 MB (1.75× source) | 155.6 MB (2.75×) | 1.57× |

The streaming reader itself is nearly free (446-byte high-water). The ~1.5–1.8× streaming floor is the metrics accumulator's proportional retention plus parse transients; materializing first stacks one full source copy on top (~1.55× the streaming peak, 57 MB extra at 50 MiB). Both paths publish identical evidence apart from `provenance.sourceKind` — asserted in the bench.

Combined with the timing side (`materialization_10MiB`: ~6% slower inline), streaming wins on both axes; the margin is memory, not time.

daveslutzkin (approved):

codesoda:
Follow-up probe (scratch run, not committed): scaled the peak-heap comparison to a ~543 MiB synthetic transcript (1.72M turns) and split the accumulator cost.

| Measurement | Value |
|---|---|
| Metrics-only peak | 797 MB (1.40× source) |
| Composite (metrics + evidence) peak | 797 MB — delta vs metrics-only is −92 KB, i.e. noise |
| Metrics retained at end of stream | 525.6 MB (~305 bytes/turn, ≈0.92× source) |
| Serialized evidence | 4,497 bytes — flat from 1 MiB to 500 MiB |

Two conclusions worth recording:

1. **The peak multiplier improves with scale**: 1.75–1.84× source at 10–50 MiB drops to 1.40× at 500 MiB. Parse transients amortize; what remains is metrics retention, linear at ~305 B/turn.
2. **The accumulator peak is entirely session metrics.** Evidence contributes nothing measurable — its peak sits inside the noise floor and its output is a flat ~4.5 KB regardless of session size. The evidence caps (`EVIDENCE_STRING_CAP`, `MAX_EVIDENCE_EXAMPLES`, …) act as memory bounds as well as privacy bounds. Metrics retains per-turn data by design (time-bucket and per-model folds).

So the pipeline memory story in one line: **peak ≈ metrics retention (~0.9–1.0× source) + bounded transients; reader ~446 bytes; evidence ~free.** If field sessions ever reach this scale and cause pressure, the lever is incremental folding of metrics retention — not the reader, not evidence. For context, 500 MiB is ~10× the largest session observed in the #222 field cohort.

codesoda:
The 500 MiB probe from the previous comment is now committed as a permanent tier in `memory_baseline` rather than a scratch run — the corpus generator materializes it on demand, so the repository stores no large blob (benches are compile-checked in CI via `clippy --all-targets` but only executed manually). BASELINE.md now records the accumulator split: composite peak 797 MB (1.40× source), of which evidence contributes 5,825 bytes.

Issue #236, opened 2026-08-27
Reduce shell memory held by the system allocator after transcript parsing


daveslutzkin:
## Summary

The shell process holds about 150 MB in a release build. Only about 20 MB of that is live data. The rest is freed memory that the macOS system allocator keeps as dirty pages.

## Evidence

Measured on the `v0.1.0-rc.8` prerelease build and on a debug build with `heap` and `vmmap --summary`:

- `antiburn` (shell): 146 MB physical footprint after 20 popover open and close cycles.
- Live heap: about 22 MB.
- `MALLOC_LARGE_REUSABLE`: about 100 MB dirty.
- `MALLOC_MEDIUM` (empty): about 115 MB reserved.

The transient buffers come from transcript parsing. The parser reads a full transcript into a `String` or a `Vec`, releases it, and the allocator keeps the pages.

## Budget

The app target is less than 300 MB in total, with 400 MB as the outer limit. The current total is about 360 MB. This item is the largest single reduction that is available.

## Proposed work

1. Change the global allocator to `mimalloc` or `jemalloc`. These allocators return large freed pages to the operating system sooner. This is a small change. Measure it first.
2. If the allocator change does not reach the target, stream the transcript parse so that the large transient buffers do not exist.

## Acceptance

- The shell footprint after a full scan is less than 80 MB on a release build.
- The measurement comes from the memory probe. Run `node scripts/mem-report.mjs --release` on a build that contains the probe and compare the `shell` column before and after.
- The change adds no dead-code or deprecated-code lint suppression.

## Related

- #237: the popover webview heap.

PR #247, opened 2026-08-27
perf: bound session metrics memory

### LINKED ISSUE #245: Store parsed transcript turns as SQLite rows and remove the O(n) metrics buffer
## Status (2026-08-28)

PR #247 implements step 4 of the plan below. It removes the `turns` vector and replaces the metrics sink with a bounded reducer. On the 500 MiB tier the metrics peak falls from 797 MB to 220.6 MB, and the retained metrics state falls from 525.6 MB to 242 KB. The desktop output does not change. Read `docs/plans/issue-245-bounded-session-metrics.md` on that branch for the design.

PR #247 defers decisions 1, 2, 5 (the rebuild half), and 8. Its plan document describes the SQLite rows as "deferred, no shipped consumer demonstrates the need". That deferral is a sequencing choice, not a reversal of the decisions in this issue. The decisions below stand. The reasons are:

- **The pipeline is still O(n) after #247.** The composite peak is 317.8 MB (0.58x source). `ClaudeStreamState::max_usage_by_message_id` holds 210 MB and `SessionEvidenceAccumulator::seen_thread_uuids` holds 97 MB. Rows resolve both: the message-id dedup becomes an upsert, and the uuid lookup becomes a query. A cap on `seen_thread_uuids` must downgrade coverage to `Partial`, so a cap is not a full answer.
- **Transcripts are pruned.** Claude Code deletes transcripts after about 30 days. A feature that needs data that the current aggregates do not hold can only get that data from rows that we stored at the time.
- **The CPU argument for incremental parse is weak, and we accept that.** `BASELINE.md` shows one full reparse costs 424 ms at 50 MiB. Incremental parse is not a CPU fix. It is the way rows get written once, and it is what keeps the residual maps out of memory.

Several structures in #247 approximate a sort by timestamp with bounded state: `ReorderWindow` (64 records), `ActiveSegments` (1,024 segments), and the efficiency open-message window (64 messages). Rows with an index give that order exactly. Steps 1 to 3 and 5 should evaluate whether each of those structures can be removed once rows exist.

Behaviour changes that #247 introduces and that this issue accepts:

- Sidechain turns inside a parent transcript count as subagent turns (plan section 7.2, option b).
- Bucket positions drift above 540 metric-bearing turns. Totals stay exact.
- The five merge divergences in plan section 7.4.
- Collection caps (`skill_uses` 256, models 32, tool names 256, and others in `metrics_sink/tally.rs`). Truncation counters exist but do not reach the UI. A later change should surface truncation through coverage.

## Summary

Store each parsed transcript turn as a row in the local SQLite database. Parse each transcript incrementally in the background worker. Compute evidence, chart buckets, and efficiency as aggregates that the pipeline updates incrementally and can rebuild from the rows. Remove the in-memory `turns` vector from the metrics sink.

This issue records the analysis and the decisions. It does not give the full implementation detail. Read it before you change the analysis pipeline, so that you do not repeat this discussion.

## Problem

The session analysis holds memory that grows with the transcript size. A measurement on a 543 MiB synthetic transcript (1.72M turns) gave:

| Measurement | Value |
|---|---|
| Metrics-only peak | 797 MB (1.40x source) |
| Composite (metrics + evidence) peak | 797 MB, no measurable difference |
| Metrics retained at end | 525.6 MB (~305 bytes/turn) |
| Serialized evidence | 4,497 bytes (same as a 1 MiB session) |

antiburn is an always-running background utility. Memory and CPU are the first two constraints. Disk is a distant third.

## Where the growth is

The only field that grows with the turn count is `SessionMetricsAccumulator.turns: Vec<MetricTurn>` in `crates/antiburn-local/src/analysis/metrics_sink.rs`. The accumulator keeps every turn until `finalize_metrics` runs at the end of the file. Three consumers need the full history:

1. **The context chart buckets** (`BUCKETS = 180`). A turn goes into the bucket for its fraction of the total active time. The total is not known until the last event. This is the structural reason for the buffer.
2. **`thread_efficiency_from_inputs`** (`efficiency.rs`). It sorts all turns by timestamp before a forward pass. The sort is defensive. The forward pass itself needs only the previous turn.
3. **`merge_metrics`**. It sorts the parent and subagent turns together by timestamp.

The other work in `finalize_metrics` (model breakdown, tool tallies, the primary cache-rehydration pass) is already written as running state. It reads the buffer only because the buffer exists.

The evidence sink (`evidence_sink.rs`) is bounded by `MAX_*` caps, with one exception: `seen_thread_uuids: HashSet<String>` has no cap and grows by one entry per record. The synthetic corpus in `tests/support/corpus.rs` emits no `uuid` or `parentUuid`, so the memory benchmark did not exercise this set. A real Claude transcript carries a `uuid` on every record.

## What the hygiene checks need

The six session hygiene checks (session overdepth, model overthinking, overpowered subagents, obsolete model, fast mode overuse, excess cache rehydration) read `SessionEvidence` only. They never read `SessionMetrics`. Each check is computable with bounded state: counters, running min/max, last-seen slots, or capped collections. This matches the equivalent checks in the cadence repository.

## Parse cost

One streaming pass runs at ~120 MB/s on a single thread (Apple M1 Max, release build). JSON parse and normalize is ~92% of the cost. The sinks are ~8%. The cost is linear in bytes at every size from 1 MiB to 543 MiB. The expensive operation is a re-parse of bytes that the pipeline has seen before, not the accumulation.

### A. Two passes over the file
Pass 1 computes the totals. Pass 2 places each turn in its final bucket. Memory is constant. Parse cost doubles. Each append to a live session changes the denominator, so the chart must rebuild from zero. Rejected.

### B. Persist accumulator state only
Make every sink resumable with bounded state, and persist `{fingerprint, byte_offset, state}` per session. The background worker parses only the new bytes. An audit of every field in `SessionEvidence` and `SessionMetrics` showed that this works, with two exceptions that need semantic changes (`late_tools` slash-command resolution and the no-timestamp bucket fallback) and several uncapped collections.

Rejected as the only storage. It freezes the feature set to what the current state can answer. A future feature that is not incremental would need a full re-parse, and Claude Code prunes transcripts after about 30 days, so the data is gone.

### C. Store parsed rows in SQLite (chosen)
Parse each turn once, in the background worker, and insert a row. Rows are the source of truth. Aggregates are caches that the pipeline rebuilds from rows when necessary.

Objections that were raised and answered:
- *Write amplification.* One row per API turn, seconds apart. One SQLite insert is ~10 µs. The cost is not measurable next to the network call that produced the turn. A large backfill at onboarding is a one-off.
- *Queries over rows are O(n) at view time.* True, but rows do not replace aggregates. The session view reads the aggregates. Rows serve rebuilds and future features.
- *Disk.* Disk is the third priority. A row without message text is ~200 bytes.

## Decisions

1. **Rows are the source of truth.** The background worker parses each transcript incrementally and inserts one row per turn into the existing SQLite store.
2. **Capture message text from day one.** Text goes into a separate `turn_content` table in the same database file, keyed by the turn rowid. The database is local to the user account, so there is no privacy concern. A later setting can turn text capture off for users who care about disk space. Capture cannot start later, because pruned transcripts are gone.
3. **Keep the hot table narrow.** SQLite reads whole pages. A scan over a narrow `turns` table touches few pages. Put any large column last in the row. Analytics queries never join `turn_content`.
4. **One database file.** A separate archive file gives lifecycle control but no query benefit, and cross-file transactions are not atomic in WAL mode. The content table can move to an attached file later if a lifecycle need appears.
5. **Aggregates are incremental and rebuildable.** Evidence, chart buckets, and efficiency totals update per turn from bounded state, and any of them can be rebuilt from rows with a query.
6. **Chart buckets use a time quantum with merge-on-overflow.** Bucket by a fixed active-time quantum. When the count passes 360, merge adjacent pairs and double the quantum. Bucket sums, maxima, and booleans merge without loss. The `secs_since_prior_turn` and `is_cache_rehydration` pair needs an order-aware merge rule. Between merges, a live chart only grows on the right. Each stream (parent, subagent) buckets on its own clock, so `merge_metrics` becomes an elementwise add and needs no sort.
7. **Remove the `turns` vector.** Nothing keeps a whole session in memory.
8. **Incremental parse.** Persist the byte offset and a fingerprint per transcript. On each change, verify the append-only guarantee (`AppendOnlyGuarantee`, `claim.fingerprint`), parse from the offset, and checkpoint the adapter state. On a mismatch, restart that session from zero.

## Consequences for known non-incremental cases

- `late_tools` (slash-command to skill resolution in `vendors/claude.rs`) becomes an `UPDATE` on the earlier row.
- Message-id usage dedup (`max_usage_by_message_id`) becomes an upsert keyed on `message_id`. The contiguity question goes away.
- `seen_thread_uuids` becomes a lookup against the rows, or a bounded recent window.
- `skill_uses[].progress` stores the numerator; the view divides by the final active time.
- `skill_uses[].duration_ms` needs a one-event lookahead, or a row update when the next turn arrives.

## Full-parse paths that remain

A full parse still happens in these cases. Each needs a CPU budget (throttle by bytes per second, newest sessions first) rather than removal:

1. Onboarding, for transcripts that exist before install.
2. An `ANALYZER_REVISION` bump. Keep stale aggregates visible until the rebuild completes.
3. An append-only violation for one session.

## High-level plan

1. Define the `turns` and `turn_content` schema in the store. Include: session id, thread/subagent id, ordinal, timestamps, message id, role, model, speed, thinking mode, sidechain flag, usage counters, compaction markers, tool call names.
2. Make the Claude adapter checkpointable: persist byte offset, fingerprint, and the small in-flight state. Replace the end-of-file resolutions with row updates.
3. Route the background worker (`insights_worker.rs`, `scan.rs`) through incremental parse and row insert.
4. ~~Rewrite the metrics sink as per-turn aggregate updates.~~ Done in #247. Remaining from this step: `seen_thread_uuids` and `invoked_skills` in the evidence sink, and `max_usage_by_message_id` and `pending_commands` in the Claude adapter. Resolve these through rows (step 1) rather than caps.
5. Add rebuild-from-rows for each aggregate, and use it for revision bumps.
6. Add `uuid` and `parentUuid` to the synthetic corpus so the memory benchmark exercises the evidence path. #247 adds a `thread_identity` flag, used only by the 500 MiB tier. Enable it for every tier.
7. Extend the memory and pipeline benchmarks to assert bounded peak memory for the incremental path. #247 asserts a bound on retained metrics state only (640 KiB). The full-pipeline bound remains open.

## Out of scope

- A setting to disable text capture.
- Compression of stored text.
- A separate archive database file.

codesoda:
## Summary

- replace the per-turn `SessionMetricsAccumulator` buffer with bounded derived state
- preserve compactions, cache classification, efficiency, skills, model transitions, active-time placement, and parent/sub-agent merging through order-aware reducers
- keep the existing `SessionMetrics` and 180-bucket presentation contract while adding a retained-state ceiling and differential reference tests
- correct delegated/sidechain turns in parent transcripts so they contribute to sub-agent totals
- document and benchmark the memory, throughput, and approximation limits

This implements the bounded-memory portion discussed in #245 without storing parsed turns in SQLite, retaining transcript content, or adding incremental parsing.

## Memory and performance

On the 500 MiB synthetic tier:

| Measurement | Before | After |
|---|---:|---:|
| Metrics-only peak | 797.0 MB | 220.6 MB |
| Metrics state retained at end | 525.6 MB | 241,599 bytes |
| Retained-state estimate vs allocator | — | 0.16% difference |

Additional results:

- 10 MiB retained metrics state: 326,591 bytes
- saturated 40k→400k turn variation: 29,376 bytes
- 10 MiB full metrics stage: about 1.20× the previous baseline
- fully disordered reducer: 1.18–1.23× the ordered reducer, below the 1.30× plan threshold

The remaining large-session peak is primarily Claude message-ID deduplication and evidence thread-identity tracking, not the metrics reducer. This PR does not claim a bounded full-pipeline peak.

## Behavior and limits

- additive totals remain exact outside documented collection caps
- long charts progressively group nearby activity
- sessions beyond 1,024 disjoint active-time intervals use approximate active duration and positions
- the analyzer revision advances so cached analysis recomputes once
- normalized serialization and the parser revision remain unchanged

## Verification

- `cd crates/antiburn-local && cargo fmt --check`
- `cd crates/antiburn-local && cargo clippy --all-targets --locked -- -D warnings`
- `cd crates/antiburn-local && cargo test --locked`
- focused characterization, pipeline-corpus, streaming-memory, disorder, compaction, cache, late-skill, sidechain, and multi-active-segment differential tests
- desktop Rust workspace formatting, clippy, and tests
- desktop TypeScript type check and 900 Vitest tests
- `pnpm run slop:all`
- `pnpm run secrets`
- `cargo bench --bench pipeline_baseline`
- `cargo bench --bench memory_baseline`, including the 500 MiB tier

## Privacy and dependencies

- no transcript rows or message content are persisted
- no network or logging path is added
- no dependencies or lockfiles change
- all new fixtures are synthetic

Related to #245 and #236. This PR intentionally does not close #245 because its archival and incremental-parsing proposal remains separate.


z0w0 (approved):

daveslutzkin:
Cross-reference: I updated #245 with a status section that records how this PR maps onto the plan. Short version: this PR lands step 4; the deferral of rows and incremental parse is sequencing, not a reversal. The two residual O(n) maps (`max_usage_by_message_id`, `seen_thread_uuids`) are the next target and #245 resolves them through rows rather than caps. The behaviour changes in plan sections 4.1, 5.8, 7.2 (option b) and 7.4 are accepted there.

PR #246, opened 2026-08-27
feat: add Pi evidence provider (#244)

codesoda:
## Summary

- add a dedicated bounded `PiAdapter` for Pi v3 JSONL sessions
- stream metrics and evidence through the claimed-source path introduced by #243
- preserve Pi token, model, thinking, tool, compaction, and timestamp semantics
- exclude inherited fork rows using the child header timestamp without reading or retaining the private parent path
- enroll existing Pi sessions through V13 and route Pi through the desktop evidence worker, reports, badges, and support gating
- add synthetic characterization fixtures, goldens, cross-layer tests, implementation plan, support docs, and both changelogs

Closes #244.

## Stacking dependency

This draft is stacked on #243. The four rebased Codex commits are patch-equivalent to PR #243 (`git range-diff` reports `=` for all four). Merge #243 first, then refresh this branch against `main` before taking this PR out of draft.

#242 is already present in this branch through `main`. Pi characterization verifies the real merged badge policy: the supported reasoning badge can be clean, capability-missing badges remain not assessed, and partial session evidence never reads clean.

## Format and coverage decisions

- The top-level Pi row timestamp controls ordering and fork ownership.
- A child excludes rows earlier than its own session header. Missing ownership timestamps fail closed with `AttributionIncomplete`.
- The adapter checks only whether `parentSession` exists. It never reads, retains, logs, or serializes its path value.
- Pi adopts #229's recommended strict structural policy locally: nested extension payload keys are inert, while direct usage, model, role, reasoning, tool, or compaction signals fail closed.
- `customType` values are never read or persisted.
- Unknown content blocks produce bounded discriminators and at most one unusable outcome per framed record.
- Global parser, analyzer, metrics, and evidence revisions stay at `3/5/1/2`; V13 enrolls existing Pi sessions without staling Claude or Codex.

## Privacy

All committed Pi fixtures are minimal synthetic records. No real `~/.pi` transcript, prompt, response, reasoning, source code, tool argument, path, identifier, fingerprint, account value, or captured machine output is included.

Evidence retains bounded structural facts only. The source reader uses the existing 8 MiB per-record framing limit and full source-validity rechecks. Pi keeps the shared exact metrics-retention contract and adds no provider-specific total-record truncation.

## Verification

- `cargo fmt --check`, `cargo clippy --all-targets --locked -- -D warnings`, and `cargo test --locked` in `crates/antiburn-local`
  - 884 unit tests passed; one pre-existing test ignored
  - 26 Pi characterization tests passed
[...]

PR #249, opened 2026-08-28
feat: add OpenCode evidence provider

z0w0:
## Summary

- add a dedicated bounded OpenCode adapter for exported JSONL and native SQLite sessions
- stream one logical message at a time, including descendant sessions, without rendering a complete in-memory transcript
- reuse one prepared part statement and one SQLite sort per message instead of compiling and sorting twice for every message
- resolve ordinary session lineage with a title point query and compare real fork candidates through bounded role-and-text streams
- validate claimed provider fingerprints inside one read-only SQLite snapshot before publishing metrics or evidence
- extract roles, timestamps, models, reasoning modes, token classes, tools, patches, compactions, and coverage diagnostics without retaining private payload text
- enroll OpenCode in the desktop evidence worker and add the V14 backfill for existing sessions
- route provider databases through the native adapter while keeping worker permit selection independent from opaque source labels
- add synthetic characterization, bounded-state, migration, worker, report, and privacy coverage

## Stacked dependencies

This draft is based on three open PRs that have not merged yet:

- #243, Codex provider adapter
- #246, Pi evidence provider, itself stacked on #243
- #247, bounded session metrics

Their patches are present in this branch because OpenCode uses the claimed-source provider seam, desktop evidence routing, and bounded metrics accumulator they introduce. Those Codex, Pi, and metrics changes are not new OpenCode scope and should remain reviewed in their existing PRs.

The current diff against `main` is therefore intentionally much larger than the OpenCode change. Merge the dependency PRs first, then rebase this branch onto `main` before taking this PR out of draft. The resulting diff should contain only OpenCode support and the integration adjustments listed below.

Two small Codex integration fixes were found while validating the stack:

- child activity-cursor changes now requeue parent evidence even when the parent fingerprint is unchanged
- fork ownership lookahead is capped at 256 rows or 1 MiB and reports partial attribution after overflow

These fixes must either land through the dependency PRs or remain when this branch is rebased.

## OpenCode source contract

OpenCode SQLite analysis opens the provider database read-only, starts one stable read transaction, verifies the claimed root-plus-descendant fingerprint, and then streams ordered messages and their parts. The adapter retains only one logical message plus bounded evidence and metrics state.

Session detail lineage does not render synthetic JSONL. Ordinary sessions stop after a title point query. Fork candidates compare only root-session user and assistant text through two one-item streams. Malformed or oversized records, tied matches, and more than 100 candidates fail closed. Descendant and non-text payloads do not enter lineage memory.

Unknown, malformed, incomplete, and oversized analysis records degrade coverage without exposing source text. Synthetic privacy tests cover reasoning, paths, tool input and output, patches, and compaction payloads.

## Verification

- engine `cargo test`: 959 passed, 2 ignored
- engine `cargo clippy --all-targets -- -D warnings`
- desktop Rust `cargo test`: 580 passed
- desktop Rust `cargo clippy --all-targets -- -D warnings`
- desktop frontend lint and type check
- desktop frontend tests: 88 files, 900 tests passed
- desktop frontend production build
- `cargo fmt --check` in both Rust workspaces
- `pnpm run slop:all` and `aislop scan --changes`: 100/100, no findings
- `pnpm run secrets`
- `git diff --check`

## Privacy and performance

All fixtures are synthetic. The change adds no network path and persists no transcript text. SQLite traversal, record framing, lineage matching, metrics state, evidence collections, and retained names all have explicit bounds.

The production-index `provider_db_stream/visit_composite/20000_rows` benchmark improved from about 342 ms to 77 ms, or 4.4x. The equivalent 2,000-row case improved from about 32.7 ms to 7.15 ms. Both revisions used the same production index fixture. Existing memory probes remained within their prior bounds.


codesoda:
FYI @z0w0 the streaming sqlite approach looks great

codesoda (approved):

PR #250, opened 2026-08-28
fix: recognize housekeeping records and make the session check verdict self-consistent

songcarver:
## User impact
Screenshot 2026-08-28 at 12-41-56

### Most "not assessed" checks were a parser bug, not missing data

Session rows were showing verdicts like `0/1 burn check · 2 not assessed`. The UI
copy was part of the problem, but the root cause sat in the engine.

Current Claude Code builds write housekeeping records that the JSONL parser did
not recognise: `custom-title`, `bridge-session`, and `artifact-comment-monitor`.
None of them carries usage data — no tokens, no models, no timing. They are a
title string, bridge metadata, and an artifact-watch marker.

The parser treated each as an unrecognised record type, which raises
`PartialReason::UnrecognizedRecordType` and marks the **whole transcript**
partially read. A partially read session can still prove a *finding*, but the
engine deliberately refuses to call a check *clean* without complete coverage
(`crates/antiburn-local/src/insights/badges.rs`). So one bookkeeping line
anywhere in a transcript downgraded all three of that session's checks to
"not assessed".

A single `custom-title` line was enough to blank a whole session's verdict.

**Scale, measured on one developer machine.** Of the Claude Code transcripts
inside the 14-day discovery window, **95% contain at least one of the three
record types**, so 95% of recent sessions could never report a clean check
whatever their actual token behaviour. `custom-title` accounts for nearly all of
them. After this branch, the same local evidence store reports 304 sessions at
complete coverage against 2 partial, down from 306 partial.

The 2 that remain are not transcripts at all — they are `.json` sidecar files
that session discovery admits alongside real transcripts. That is a separate
discovery bug and is not fixed here.

`PARSER_REVISION` goes 3 → 5, so every stored verdict recomputes on next launch.

### The verdict line contradicted itself

`0/1 burn check · 2 not assessed` makes the reader add 1 and 2 to learn the
session has three checks, and the two numbers look like they disagree. Every
session always has exactly three checks — the denominator was counting only the
*assessed* ones.

| Data | Before | After |
| --- | --- | --- |
| 3 assessed, all clean | `3/3 burn checks` | unchanged |
| 1 assessed (failed), 2 not | `0/1 burn checks · 2 not assessed` | `0/3 burn checks · 2 not assessed` |
| 0 assessed | `0/0 burn checks · 3 not assessed` | `Not assessed` (tertiary ink, no fraction) |
| Evidence pending | `Computing checks` | `Computing checks…` |

1. **The denominator is every check, never only the assessed ones.** This also
   retires the singular-noun problem by construction: the denominator is the
   badge count, which is always three, so `0/1` can no longer appear.
2. **Never render `0/0`.** With nothing assessed the fraction is noise, so the
   verdict becomes `Not assessed` in `label-tertiary`.
3. **The ink ramps against the total too.** One failure beside two unassessed
   checks now reads orange (33%), not full red. Claiming maximum severity when
   two thirds of the checks never ran overstated the finding.
4. **The tooltip explains the grey rows.** Each not-assessed check names its
   reason in plain terms. Previously the tooltip showed a title and a `–` mark
   with no reason at all.
5. **An ellipsis marks transient states only.** `Computing checks…`,
   `Refreshing checks…`, and `Still writing checks…` end on their own; settled
   states such as `Unsupported checks` stay bare.

### Documentation

`docs/plans/session-check-states.md` records when the checks actually run: the
60-second popover-gated scan tick, the 14-day discovery window, the evidence
worker's backoff, and why a live session holds at "not assessed" until its
transcript settles. That was the question behind the original report and it was
not written down anywhere.

## Screenshot

The tooltip's new reason text, above a row reading `1/3 burn checks · 1 not
assessed` — the fixed denominator, which previously would have read `1/2`.

## Validation

- Engine (`crates/antiburn-local`): `cargo fmt --check`, `cargo clippy
  --all-targets -- -D warnings`, `cargo test` — 962 pass, 0 fail.
- Desktop shell (`apps/desktop/src-tauri`): `cargo fmt --check`, `cargo clippy
  --all-targets -- -D warnings`, `cargo test` — 566 pass, 0 fail.
- Desktop UI: `pnpm --filter @antiburn/desktop lint`, `type-check`, and `test` —
  902 pass across 88 files, eslint and `tsc` clean.
- The housekeeping characterization fixture now covers all three new record
  types, so the recognised-eventless list stays locked by a test. It covered
  none of them before.
- Manually verified against a running debug build: after the revision bump the
  local evidence store reprocessed every session and coverage moved as described
  above.

## Privacy and performance

No change to data access, network access, or retained data. The parser change
only widens the set of record types recognised as carrying no usage data, so it
reduces work rather than adding any: recognised housekeeping records short-circuit
instead of raising a partial-coverage reason.

The `PARSER_REVISION` bump makes existing users recompute stored evidence once on
next launch. That is the established mechanism for a parser change and it is
bounded by the existing worker limits (newest first, 60 analyses per pass, 30s to
15m backoff).

## Checklist

- [x] I used synthetic test data.
- [x] I did not include credentials, private session content, real transcripts, repository names, or private file paths.
- [x] My commits include a DCO sign-off (`git commit -s`).
- [x] I updated tests and documentation where needed.

---

## Additional functionality included

This PR also includes the session-check expansion from the Dave investigation:

- Expands session row and detail hygiene badges from the original three to six:
  - Session overdepth
  - Model overthinking (renamed from reasoning overkill)
  - Overpowered subagents
  - Obsolete model
  - Fast mode overuse
  - Excess cache rehydration
- Keeps cache rehydration in the set; it was crossed out in the original checklist because it was already implemented.
- Wires the new badge identifiers through the Rust projection, Tauri DTO, TypeScript IPC and presentation layers, and UI tests.
- Reuses evidence already accumulated by the engine. This addition exposes existing detector results at session level rather than adding a separate collection path.
- The verdict-state improvements above now apply consistently across all six checks.


songcarver:
@codesoda you prob have thoughts

songcarver:
Now with fix 
Screenshot 2026-08-28 at 13-04-20

codesoda (approved):

PR #271, opened 2026-08-30
feat(opencode): stream descendant sessions as delegated threads

daveslutzkin:
Seam 4e of `docs/plans/session-evidence-harness-parity.md`. Follows #270.

## What changes

- Every OpenCode message row gets `thread_id` = its own session id.
- A `parent_id` descendant's messages stream as `EventSource::Subagent` → `Delegated` scope, `child_id` = session id. Subagent turns no longer fold into the parent's transitions, idle gaps, or model set.
- One `SubagentSpawn` per descendant session (new `RelationProvenance::SessionParentLink`), `ts_ms` from the child session's `time_created`, `parent_model` = the root's current model. Both the SQLite path (one-shot `session` join over the cluster) and the export path (`session_member` rows keyed by `originSessionID`, `message.sessionRole == "child"` / `sessionID != root`) apply the same rule.
- Tracked children capped at 50 000 like `ThreadResolver`; past the cap → `AttributionIncomplete`.
- A `parent_id` child whose title matches `opencode_fork_parent_title` is still delegated but degrades `subagents` to `Partial(AttributionIncomplete)`. Forks (null `parent_id`) never enter the cluster; fixtures freeze that. Fork detection stays in discovery.
- `SourceCapabilities::opencode()`: `subagent_relationships`, `subagent_models`, `thread_identity` → true. `SessionsOverDepth`, `OverpoweredSubagents`, and `CacheChurn` become assessable for OpenCode; `OveruseOfFastMode` stays out (no speed signal).
- `PARSER_REVISION` 8→9, `EVIDENCE_SCHEMA_REVISION` 6→7. The 8 golden diffs (4 Codex, 4 Pi) are revision-only.

## Tests

New in `opencode_characterization.rs`: delegated child + grandchild threads, fork is its own root, fork-title-without-prefix is an ordinary root, fork-titled `parent_id` child degrades attribution, export-path spawn. Desktop acceptance test `an_opencode_parent_id_child_links_as_a_delegated_thread`. Test DB schemas gained the `title` column the new query joins.

fmt / clippy `-D warnings` / tests clean in both crates.

### DIFF
--- a/apps/desktop/src-tauri/src/analysis.rs
+++ b/apps/desktop/src-tauri/src/analysis.rs
@@ -1651,7 +1651,8 @@ mod tests {
         connection
             .execute_batch(
                 r#"CREATE TABLE session (
-                     id TEXT PRIMARY KEY, parent_id TEXT, time_created INTEGER, time_updated INTEGER
+                     id TEXT PRIMARY KEY, parent_id TEXT, title TEXT,
+                     time_created INTEGER, time_updated INTEGER
                  );
                  CREATE TABLE message (
                      id TEXT PRIMARY KEY, session_id TEXT, time_created INTEGER,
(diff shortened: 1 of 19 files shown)


daveslutzkin:
Follow-up commit: the sink treats a null `uuid` as a thread-identity gap once a source claims `thread_identity`, which silently degraded every OpenCode `cache` group to `Partial(AttributionIncomplete)`. The adapter now carries the message id as `uuid` on both paths; the delegated-threads test asserts `cache` is `Complete`.

PR #281, opened 2026-08-30
fix(claude): skip fork-replayed subagent rows at ingest

daveslutzkin:
## Problem

Newer Claude Code harnesses support **fork** subagents. A fork inherits the parent
agent's full conversation, and its transcript file replays the parent agent's
records with the SAME uuids before appending its own records (which have new
uuids).

Layout on disk (a real session showed 389 of a parent's 529 records replayed
into its fork's file):

```
<session-id>/subagents/agent-aaaa000000000001.jsonl        # a normal subagent
<session-id>/subagents/agent-aaaa000000000001.meta.json
<session-id>/subagents/agent-bbbb000000000002.jsonl        # a fork of it: replays
<session-id>/subagents/agent-bbbb000000000002.meta.json    # agent aaaa's rows first
```

Effects in antiburn today:
- The replayed rows land in the `turn` table twice (once per source file), so
  `query_duplicate_turn_identities` (`crates/antiburn-local/src/analysis/evidence_query.rs`)
  reports hundreds of duplicates.
- `evidence_sink.rs` then degrades the `models` and `subagents` groups to
  `Partial(attribution_incomplete)`, so four of the six burn checks show
  "not assessed" for any session that used a fork agent.
- The replayed rows carry usage, so token and spend aggregates double-count
  (a real session double-counted ~29.8M cache-read tokens).

## Fix

At ingest, when reading a subagent transcript whose sibling `meta.json` has
`"isFork": true`:

1. Determine the replay source: the direct parent transcript — the sibling
   `agent-<parentAgentId>.jsonl` when `parentAgentId` is present, else the
   session's main transcript (a fork of the top-level session). A chain of
   forks needs only the direct parent: the parent file already contains its
   own replayed rows, so the direct parent's uuid set covers the chain
   transitively.
2. Collect the uuid set of that replay source file and skip records whose
   uuid is in the set: write no turn row and no usage for them.
3. Fail open: a missing or unreadable meta.json or parent file means skip
   nothing — keep today's behavior (duplicates degrade to
   `attribution_incomplete`; the session never becomes wrongly Clean).

The skip sits in `analysis/vendors/claude.rs`'s parse loop, before a record's
`uuid` can become a turn row through `TurnRowSink` — the ingest boundary, not
the duplicate-identity query or the sink's degrade logic, which stay
unchanged and remain the backstop for genuine identity corruption.

`crates/antiburn-local/src/discovery/mod.rs`'s `SubagentMeta` gains `is_fork`
(`isFork`) and `parent_agent_id` (`parentAgentId`) fields.

`PARSER_REVISION` bumps 12 → 13 (`crates/antiburn-local/src/analysis/mod.rs`)
so stored fork sessions re-ingest and recover. Two mirrored test literals
pinning `parserRevision: 12` (`crates/antiburn-local/src/analysis/evidence.rs`
and `apps/desktop/src-tauri/src/store/tests.rs`) are updated to 13, along with
the codex/pi characterization goldens that embed `provenance.parserRevision`
(regenerated with `UPDATE_GOLDENS=1`; each diff is exactly that one field).

**Note:** a parallel PR (codex token_count heartbeat) also bumps
`PARSER_REVISION`. Whichever merges second must rebase and re-bump.

## Tests

- `crates/antiburn-local/tests/claude_characterization.rs`:
  `fork_replayed_uuids_are_counted_once_and_do_not_degrade_evidence` streams a
  parent, a normal subagent, and a fork of that subagent (new
  `fork_replay_*.jsonl` / `.meta.json` fixtures, real files on disk under a
  `subagents/` directory) through the same shared-store ingest shape
  production uses, and asserts: the replayed uuid appears exactly once in the
  `turn` table under the normal subagent's own `source_key`; its usage is
  counted once; `duplicate_turn_identities` is zero; `models` and `subagents`
  stay `Complete`.
- `crates/antiburn-local/src/analysis/vendors/claude.rs` unit tests cover the
  meta.json reader and path resolution: fork with `parentAgentId`, fork
  without it (parent = main transcript), malformed JSON (treated as
  non-fork), and fail-open when the parent file is missing — proving nothing
  is skipped, i.e. today's duplicate/degrade behavior is preserved. A
  `visit_reader`-level test proves a replayed record is dropped while a new
  one survives.

## Checks

Both crates (`crates/antiburn-local`, `apps/desktop/src-tauri`): `cargo fmt`,
`cargo clippy --all-targets -- -D warnings`,
`UPDATE_GOLDENS=1 cargo test --no-fail-fast`, then `cargo test --no-fail-fast`.
13 test-result lines in antiburn-local, 3 in src-tauri, all `ok`.


daveslutzkin:
Rebased onto main after #280 merged: `PARSER_REVISION` re-bumped 13 → 14 (both fixes re-ingest stored sessions), mirrored test literals and goldens regenerated, full checks green in both crates.

PR #290, opened 2026-08-30
feat(codex): read cache-write tokens into usage

daveslutzkin:
## Summary

- Current Codex CLIs write `cache_write_input_tokens` on every `token_count` usage object, and OpenAI now bills cache writes for the GPT-5.6 family (`pricing/table.rs` already has nonzero `cache_write_cost_per_token` for sol/terra/luna). `codex_usage` now mirrors Cadence's `parse_codex_last_token_usage` (`crates/analysis/src/model_turns.rs`): it takes the largest of the cache-read alias keys and the largest of the cache-write alias keys, subtracts both from `input_tokens` (floored at 0), and populates `cache_creation_tokens` instead of dropping it.
- The Codex capability `cache_write_tokens_available` now starts false and becomes sticky-true once an owned `token_count` usage object carries a cache-write alias key (`usage_carries_cache_write`), on both Codex read paths (`normalize`/`parse_codex` and the streaming `visit`/`visit_claimed` path via `CodexStreamState`).
- Codex's repeated-context accounting stays pinned to uncached-input even once that capability turns on, matching Cadence's `CacheAccountingMode::for_harness`: the overpay-band bounds were derived under uncached-input accounting, so cache-write tokens still feed cost and turn depth, they just don't flip which bucket Cache Churn scores Codex on.
- Bumps `PARSER_REVISION` 14 → 15 and regenerates characterization goldens (codex + pi; movement is `parserRevision` only outside fixtures that carry a cache-write key).

## Test plan

- [x] `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, `UPDATE_GOLDENS=1 cargo test --no-fail-fast`, `cargo test --no-fail-fast` — green in both `crates/antiburn-local` (13 suites, 1120+75+33+... all `ok`) and `apps/desktop/src-tauri` (3 suites, 612 `ok`)
- [x] New unit tests: `codex_usage_splits_cache_read_and_cache_write_out_of_input`, `cache_write_split_preserves_context_occupancy` (context-occupancy invariant), `token_count_with_only_a_cache_write_reads_as_real_usage` (heartbeat no longer fails closed), `codex_confirmed_cache_write_is_read_as_a_direct_signal_rehydration` (mode 1 now reachable), `a_priced_codex_turn_with_a_cache_write_pays_the_write_premium`
- [x] New `cache_write_tokens` characterization fixture + golden, asserting the per-turn split, `capabilities.cacheWriteTokens: true`, `repeatedContext.accounting: uncached_input`, and unchanged context occupancy
- [x] Updated `record_to_event_changes_require_an_inertness_review` fingerprint after reviewing the inertness table (`CODEX_SCALAR_EVIDENCE_KEYS` now lists the cache-write/read alias keys `codex_usage` reads)
- [x] Fixed three pre-existing `analysis::tests` fixtures that explicitly carried `"cache_write_input_tokens":0` (now that presence of the key sets the capability true, they were unintentionally routing into the confirmed direct-signal path instead of the heuristic-inference path they exist to test) by dropping that key so they represent genuine no-signal Codex sessions

Note: a parallel PR bumps `ANALYZER_REVISION` and regenerates goldens for the fast-mode/overthinking detectors. Whichever merges second must rebase and regenerate.


daveslutzkin:
Rebased onto main after #282 merged: goldens and mirrored literals now carry parser 15 + analyzer 15; full checks green in both crates.

PR #296, opened 2026-08-31
feat: configure session data retention

z0w0:
## Summary
- add 30-day, 90-day, and forever retention choices, with forever as the default
- remove expired app-owned session records transactionally at setting changes, startup, and daily cleanup
- confirm destructive retention changes and refresh open views and repository counts after cleanup

Screenshot 2026-08-31 at 11 20 38 am

## Privacy and performance
- retention applies only to antiburn's local SQLite data and never modifies provider transcript files
- longer retention preserves history beyond providers' 30-day window
- shorter retention keeps the local index lighter

## Validation
- `pnpm --filter @antiburn/desktop lint`
- `pnpm --filter @antiburn/desktop type-check`
- `pnpm --filter @antiburn/desktop test` (939 tests)
- `pnpm --filter @antiburn/desktop build`
- `cargo fmt --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo test` (623 tests)
- `pnpm run slop:all`
- `pnpm run secrets`


apps/desktop/src-tauri/src/store/mod.rs
```diff
@@ -283,6 +281,34 @@ impl Store {
         self.update_settings(|current| *current = settings.clone())
     }
 
+    /// Replace preferences and apply their retention policy in one transaction.
+    pub fn replace_settings_and_apply_retention(
```
daveslutzkin: shouldn't this be more of a HTTP PATCH endpoint? like replace settings and apply some sort of generic change to them? otherwise we'll need specific methods for everything

daveslutzkin (approved):

apps/desktop/src-tauri/src/store/mod.rs
z0w0: Replaced the retention-specific method with a generic `replace_settings_with` transaction callback. The settings command now applies retention through that callback, so future transactional settings side effects do not require dedicated store methods.

PR #303, opened 2026-08-31
feat(desktop): widen the evidence cohort to every agent and retire the on-demand analysis paths

daveslutzkin:
## Summary

- **(a0) Make `capabilities_for_vendor` total.** `capabilities_for_vendor` (apps/desktop/src-tauri/src/analysis.rs) covered only 4 of 11 agent vendors, returning `Option<SourceCapabilities>`. Added `SourceCapabilities::cursor()`, `::antigravity()`, and `::generic()` to `crates/antiburn-local/src/analysis/evidence.rs`, each derived by reading the actual adapter code (not copied from another vendor), and made the function total (`SourceCapabilities`, no `Option`). This let every vendor take the normal evidence-streaming path; the engine's existing `published_status` classifies evidence that satisfies no detector as `PublishedEvidence::Unsupported`, which is the honest terminal state for an uncharacterized vendor. Deleted the now-unreachable legacy `analyze_sources_with` branch inside `analyze_for_evidence`.
- **(a) Widen the evidence cohort.** `evidence_cohort()` (apps/desktop/src-tauri/src/agents.rs) now derives from `AgentKind::ALL` (11 kinds) instead of a hand-maintained 4-agent array, so it cannot drift from the engine's discovery matrix again. Every agent's sessions now flow through the durable evidence worker and turn-rows pipeline. Added migration V20 (`apps/desktop/src-tauri/src/store/schema.rs`) backfilling pending `session_evidence` rows for existing sessions of the seven newly-joining agents (Cursor, Copilot, Cline, Kiro, Amp, Antigravity, Windsurf), following the V12/V13/V14 template; it does not reset or clobber an existing evidence row.
- **(b) Retire the legacy on-demand analysis paths.** With the cohort universal, `cache_detail_analysis`, `analysis_changed`, `cache_detail_relations`, `top_up_analysis`/`TopUpScope`/`MAX_ANALYSES_PER_PASS`, and their exclusive plumbing are dead. Deleted them. `get_session_analysis`/`get_subagent_analysis` now unconditionally serve the rows-first path with a live-parse fallback. `Store::save_analysis` is demoted to `#[cfg(test)]` test scaffolding (its only remaining callers are fixture setups); the three tests that pinned its column writes are re-pointed at `publish_projections`, the worker's real write path, so column coverage is not lost.

## Consequences / notes for reviewers

- **`awaiting_provider_support`** (insights_report.rs's `DENOMINATOR_SQL`) is exactly the non-cohort population, so it trends to zero once the V20 backfill runs. New agents land in the existing `unsupported`/`ready` coverage buckets instead. The field and SQL are kept (DTO compatibility); the pinning tests were updated to reflect this rather than deleted.
- **`SESSION_ENTRY_CHANGED_EVENT` timing**: the synchronous emission from drilldown open (`cache_detail_analysis`'s call site) is gone. The worker's own announce callback (`insights_worker::spawn`) still emits this event on every publish, so the popover list still updates — just from the worker's pass rather than the drilldown's.
- **`supports_analysis`/`has_dedicated_adapter` are unchanged** (explicit non-goal). Drilldown UI gating stays on the adapter set; widening the cohort only makes evidence rows, hygiene, list cost pills, and provider-usage attribution exist for the new agents. Zero TS changes.
- **Known limitation**: Cursor, Antigravity, and the generic-JSONL kinds have no characterization fixture corpora in this repo. Their worker path is covered only by the new synthetic tests (engine-side `SourceCapabilities` pins reading the real adapter code; desktop-side a synthetic generic-agent `process_next` test asserting a terminal published status, never a stuck `processing` claim).
- Base: this branch was cut from `origin/feat/drilldown-from-rows` (PR #302), which has since merged into `main`; this PR now targets `main` directly and contains only the 3 commits above.

## Test plan

Both crates, per repo convention:
- [x] `crates/antiburn-local`: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --no-fail-fast` — 14 binaries, all `0 failed` (1143 + 2 + 75 + 33 + 11 + 30 + 10 + 8 + 10 + 5 + 8 + 5 + 11 + 1 passed)
- [x] `apps/desktop/src-tauri`: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --no-fail-fast` — 3 binaries, all `0 failed` (645 passed)
- [x] Zero changes under `apps/desktop/src/**` (confirmed via diff against `origin/main`)

### DIFF
--- a/apps/desktop/src-tauri/src/agents.rs
+++ b/apps/desktop/src-tauri/src/agents.rs
@@ -44,13 +44,11 @@ pub fn kind_from_slug(slug: &str) -> Option<AgentKind> {
 }
 
 /// Return the agents that use the durable evidence queue.
-pub fn evidence_cohort() -> [&'static str; 4] {
-    [
-        AgentKind::Claude.slug(),
-        AgentKind::Codex.slug(),
-        AgentKind::OpenCode.slug(),
-        AgentKind::Pi.slug(),
-    ]
+///
+/// Every supported [`AgentKind`] is in the cohort, so this can never drift
+/// from the engine's discovery matrix.
+pub fn evidence_cohort() -> Vec<&'static str> {
+    AgentKind::ALL.iter().map(|kind| kind.slug()).collect()
 }
 
 #[cfg(test)]
@@ -106,13 +104,21 @@ mod tests {
     fn the_evidence_cohort_uses_the_discovery_slug() {
         assert_eq!(
             evidence_cohort(),
-            [
-                AgentKind::Claude.slug(),
-                AgentKind::Codex.slug(),
-                AgentKind::OpenCode.slug(),
-                AgentKind::Pi.slug(),
+            vec![
+                "claude-code",
+                "codex",
+                "cursor",
+                "copilot",
+                "cline",
+                "opencode",
+                "kiro",
+                "amp-code",
+                "antigravity",
+                "windsurf",
+                "pi",
             ]
         );
+        assert_eq!(evidence_cohort().len(), AgentKind::ALL.len());
         assert_eq!(AgentKind::Codex.slug(), vendor_label(AgentKind::Codex));
         assert_eq!(
             AgentKind::OpenCode.slug(),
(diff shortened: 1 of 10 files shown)


daveslutzkin:
Dropped `widened_cohort_migration_queues_existing_sessions_without_resetting_evidence` per maintainer review: it pins a one-shot property of a frozen migration, and the failure it guards against (resetting an existing evidence row) is self-healing — the worker would just reprocess the session. Desktop suite green after removal (644 passed, 0 failed).

PR #346, opened 2026-09-02
feat(desktop): restyle the Session Detail view

songcarver:
## User impact

The Session Detail view read as a different product from the popover home screen: six type styles, thirteen ink utilities, and three unrelated bar treatments. This pass puts it on the home screen's rules and splits the body into three tabs: Context, Cost, Tools.

Screenshot 2026-09-04 at 22-38-30
Screenshot 2026-09-04 at 22-39-27
Screenshot 2026-09-04 at 22-39-37

Screenshot 2026-09-04 at 22-39-59
Screenshot 2026-09-04 at 22-40-11

Screenshot 2026-09-04 at 22-40-24

**Type and labels**

- One data size per surface. Every figure and row uses `type-body`; hierarchy comes from ink and weight, not from a size change. The ladder drops from six styles to four.
- Section headings that restate their tab are gone, and so are the caption labels over self-evident values. Identification moves to icons with tooltips, the pattern the session rows already use.
- Hygiene checks read as a plain verdict word, `Passed` or `Failed`, with the icon after it. Only a failure takes colour. The row explainer sits in the gutter on hover, so the verdicts reach the right margin.

**Context tab**

- The context chart fills the tab's height. Token series paint as solid fills over the context area, and every text label renders after the plot layers, because SVG has no z-index.
- The chart plays in as a sequence: context, then tokens, then the rewrite marks.
- The work, rewrite and carry shares draw as one composition track under the chart key, since they are parts of one whole. Real work takes the label ink, rewrite waste the brand orange, and carry a mid neutral. Each row's guidance sits in a tooltip so the block keeps the height of its readings.

**Cost tab**

- Checks lead, then the cost rows, then Efficiency.
- `$/MTok` draws as a bullet graph: three grey bands for good, ok and high, with the word and dollar range under each, the reading as a brand-orange measure, and a target mark at the good edge. No traffic-light colour anywhere: the words carry the verdict.
- The reading's guidance prints inline under the scale, and the Efficiency section sits at the foot of the tab, so a short tab keeps its slack between the cost rows and the scale rather than after it.

**Tools tab**

- The headline states the count of items loaded but never called, on one line with its icon, and the subline says what the number is.

**Known limits**

- `.ui-tooltip` now paints on the `surface` token instead of two named greys. That is the surface the chart tooltips already used, but it changes every platform tooltip in the app, not only these.
- The `share-work` and `share-waste` tokens are now used only by the hygiene verdicts. Their comments in `design.md` still name the efficiency composition and could be renamed in a follow-up.

## Validation

Run from `apps/desktop` unless noted.

- `pnpm format:write` on every touched file
- `pnpm type-check` — clean
- `pnpm lint` — clean
- `pnpm knip` — clean
- `pnpm test` — 1145 passed, 93 files
- `node scripts/check-design-drift.mjs` (repo root) — design contract in sync with the stylesheets

New tokens are documented in `apps/desktop/design.md` in the same change, as the repository contract requires. The review rounds are recorded in `docs/plans/session-detail-style-guide.md` and `docs/plans/session-detail-playground-build.md`.

## Privacy and performance

None. This change is presentation only: no new data access, no network access, no retained data, and no background work.

## Checklist

- [x] I used synthetic test data.
- [x] I did not include credentials, private session content, real transcripts, repository names, or private file paths.
- [x] My commits include a DCO sign-off (`git commit -s`).
- [x] I updated tests and documentation where needed.


z0w0 (approved):

z0w0:
Screenshot 2026-09-07 at 9 09 47 am

Should these be different colors? Green vs red

PR #359, opened 2026-09-03
Fix AppImage startup on newer Linux graphics stacks

locusrifle:
## What this fixes

The Linux AppImage includes an old graphics library that conflicts with newer Linux systems. Antiburn starts, but its interface never appears. This change makes the AppImage use the computer's compatible library instead and adds a test to prevent the problem from returning.

## How

- Pin a newer `linuxdeploy` build by GitHub asset ID and SHA-256.
- Verify every release AppImage excludes `usr/lib/libwayland-client.so.0`.
- Keep the Debian package, macOS and Windows release paths unchanged.

Tauri 2.11.4 otherwise downloads its cached 2024 `linuxdeploy` build. The pinned build follows the AppImage graphics-library exclusions, so host graphics drivers resolve the host Wayland client ABI.

## Reproduction and verification

I reproduced this with the checksum-verified v0.3.2 AppImage on Arch Linux under Wayland and Hyprland. The stock AppImage emitted `EGL_BAD_PARAMETER`, and its WebKit renderer did not become ready. Removing only the bundled Wayland client fixed startup.

I then built the AppImage from this branch with the pinned bundler. The artifact passed the new package boundary check and its onboarding renderer became ready successfully.

Checks run:

- 94 CI control-plane tests
- `pnpm run secrets`
- `pnpm run slop` — score 100
- Local Tauri debug AppImage build and launch
- Boundary check against both the failing official artifact and the corrected artifact

## Privacy and performance

This changes only the Linux release packaging path. It does not read or transmit credentials, sessions, analytics, or other user data. Application runtime behavior and resource use are unchanged. The Linux release job downloads one checksum-pinned bundler artifact before building.

### DIFF
--- a/.github/workflows/release-app.yml
+++ b/.github/workflows/release-app.yml
@@ -581,6 +581,26 @@ jobs:
           test -f apps/desktop/src-tauri/icons/icon.icns
           test -f apps/desktop/src-tauri/icons/icon.ico
 
+      # Tauri 2.11.4 caches a 2024 linuxdeploy build that bundles libwayland-client.
+      # This pinned build follows the AppImage graphics-library exclusions.
+      - name: Install the pinned Linux AppImage bundler
+        if: runner.os == 'Linux'
+        shell: bash
+        env:
+          LINUXDEPLOY_ASSET_ID: "538917371"
+          LINUXDEPLOY_SHA256: "36a2d7e274d12e1050d0e9ecfe11d339ed54720b2bec464c286d53f8b07f5c62"
+        run: |
+          set -euo pipefail
+          tools="${XDG_CACHE_HOME:-$HOME/.cache}/tauri"
+          destination="${tools}/linuxdeploy-x86_64.AppImage"
+          mkdir -p "$tools"
+          curl --fail --location --retry 3 \
+            --header 'Accept: application/octet-stream' \
+            "https://api.github.com/repos/linuxdeploy/linuxdeploy/releases/assets/${LINUXDEPLOY_ASSET_ID}" \
+            --output "$destination"
+          echo "${LINUXDEPLOY_SHA256}  ${destination}" | sha256sum --check --strict
+          chmod +x "$destination"
+
       - name: Build and sign
         shell: bash
         env:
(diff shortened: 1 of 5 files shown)


.github/workflows/release-app.yml
```diff
@@ -581,6 +581,26 @@ jobs:
           test -f apps/desktop/src-tauri/icons/icon.icns
           test -f apps/desktop/src-tauri/icons/icon.ico
 
+      # Tauri 2.11.4 caches a 2024 linuxdeploy build that bundles libwayland-client.
+      # This pinned build follows the AppImage graphics-library exclusions.
+      - name: Install the pinned Linux AppImage bundler
+        if: runner.os == 'Linux'
+        shell: bash
+        env:
+          LINUXDEPLOY_ASSET_ID: "538917371"
```
codesoda, line 590: **[🟡 NOTE] `linuxdeploy` Tauri 2.11.4 cached build → asset `538917371`

The pinned AppImage matches the independent SHA-256 check (`36a2d7e2…`) and comes from the unchanged `linuxdeploy/linuxdeploy` owner, but the asset is from the rolling `continuous` pre-release and has no SLSA/Sigstore attestation. Safe to merge; revisit the pin when upstream replaces the rolling build. [Asset metadata](https://api.github.com/repos/linuxdeploy/linuxdeploy/releases/assets/538917371) · [upstream repository](https://github.com/linuxdeploy/linuxdeploy)

<sub>`dependa-audit` · reply to resolve</sub>

codesoda:
## 🔍 dependa-audit — Safe to merge (notes below)

Audited 1 dependency-like release artifact introduced by this PR. It was checked for install hooks, dependency changes, publisher/provenance, network endpoints, and code diff.

| Package | From → To | Eco | Hooks | New deps | Provenance | Publisher Δ | Verdict |
|---------|-----------|-----|-------|----------|------------|-------------|---------|
| `linuxdeploy` | Tauri 2.11.4 cached build → asset `538917371` (`continuous`) | gha | fetch + SHA-256 verify | none | none | no | [🟡 NOTE](https://github.com/antiburn/antiburn/pull/359#discussion_r3919928792) |

### Summary

- No package-manager manifests or lockfiles changed, and no new npm, Cargo, or other registry packages were introduced.
- The only dependency-like change pins the existing Linux bundler to a GitHub asset ID and checks its SHA-256 before execution.
- No install scripts, secret reads, runtime changes, or suspicious network endpoints were found. The new download uses one `api.github.com` release-asset request.
- The new verification code uses Node.js built-ins only and enforces that the built AppImage excludes `usr/lib/libwayland-client.so.0`.
- The upstream owner and GitHub Actions uploader are unchanged. No SLSA/Sigstore attestation is attached.

### Worth knowing

`linuxdeploy` comes from the rolling `continuous` pre-release channel and has no SLSA/Sigstore provenance beyond the pinned digest. An independent download matched SHA-256 `36a2d7e274d12e1050d0e9ecfe11d339ed54720b2bec464c286d53f8b07f5c62`; the upstream repository remains owned by `linuxdeploy` and is unarchived. This is a maintenance note, not a merge blocker. Revisit the pin when upstream replaces the rolling build, or prefer a tagged release with attestations if one becomes available. [Asset metadata](https://api.github.com/repos/linuxdeploy/linuxdeploy/releases/assets/538917371) · [upstream repository](https://github.com/linuxdeploy/linuxdeploy)

z0w0 (approved):
Thanks for your PR!

I think the linuxdeploy release asset might need to be updated in the future if the release rolls, but we can solve that problem later - better to get the builds working well with Wayland for now.

Otherwise looks good on my end, I suspect we'll need to keep an eye on this as not sure how it'll go across distros. Would appreciate any help you can lend in the future to keep testing the Linux builds.

.github/workflows/release-app.yml
z0w0: Acceptable for now, but will need to keep an eye on for future builds

PR #379, opened 2026-09-03
feat: add aggregated checks

z0w0:
## What this changes

- Adds one "All checks" summary to the Activity popover.
- Opens a passive detail card when the reader hovers or focuses the summary.
- Groups checks into issues and passed checks, with plain explanations and an estimated token-burn share.
- Refreshes the report after scans without interrupting the current view.
- Keeps report data bounded so large session histories do not make the desktop app grow without limit.

## Why

Users can now see recurring sources of avoidable AI spend across their recent sessions without opening sessions one at a time. The summary shows only results that the app could assess and stays honest when refresh data is unavailable.

## Screenshot

Screenshot 2026-09-03 at 6 33 31 pm

## Validation

- 1,116 frontend tests passed.
- 832 desktop Rust tests passed.
- 1,172 engine unit tests and all integration suites passed.
- Formatting, lint, type checks, dead-code checks, production build, design drift, secret scanning, and Rust Clippy passed.
- The code-quality scan scored 100/100.
- The release memory report stayed below the 100 MiB limit.


apps/desktop/src/views/PopoverView.tsx
```diff
@@ -46,6 +48,9 @@ const SessionPane = lazy(() =>
 )
 
 function samePopoverPeekTarget(left: PopoverPeekTarget, right: PopoverPeekTarget): boolean {
+  if (left.kind !== right.kind) return false
+  if (left.kind === "checks" && right.kind === "checks") return true
+  if (left.kind === "checks" || right.kind === "checks") return false
```
daveslutzkin, line 53: that's some confusing logic

daveslutzkin (approved):

PR #449, opened 2026-09-08
feat(desktop): add main window and persistent Activity sidebar

martyportier:
## Stack

1. **PR #449 — main-window foundation** (this PR, targets `main`)
2. [PR #450 — collection/detail navigation](https://github.com/antiburn/antiburn/pull/450)
3. [PR #451 — Sessions integration](https://github.com/antiburn/antiburn/pull/451)

## Summary

- Add the retained desktop main-window lifecycle and dedicated renderer.
- Add native window controls, placement persistence, readiness handling, and activation behavior.
- Keep background/login launches quiet while explicit launches open the main window.
- Add the persistent Activity sidebar foundation.
- Preserve existing tray, popover, Settings, HUD, notification, and monitoring behavior.
- Add lifecycle diagnostics, CI coverage, and the main-window validation runbook.

## Platform behavior

- macOS uses native traffic lights and restores closed or minimized windows on activation.
- Windows and Linux retain native title bars and taskbar behavior.
- Login registration uses an explicit background launch argument where required.

## Validation

- Frontend formatting, lint, type-check, tests, and production build.
- Rust formatting, Clippy, and tests for the shell and main-window crate.
- Packaged macOS debug build and lifecycle smoke testing.
- Windows/Linux native release validation remains required and is documented in the runbook.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Test the CI control plane
-        run: node --test scripts/build-tool-catalog.test.mjs scripts/check-design-drift.test.mjs scripts/classify-ci-changes.test.mjs scripts/mem-report.test.mjs scripts/verify-linux-appimage.test.mjs scripts/third-party-notices.test.mjs scripts/verify-app-engine-release.test.mjs scripts/wait-for-main-ci.test.mjs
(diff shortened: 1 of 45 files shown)


.github/workflows/ci.yml
```diff
@@ -201,6 +201,10 @@ jobs:
         working-directory: apps/desktop/src-tauri/crates/hud
         run: cargo fmt --check
 
+      - name: Check main-window crate formatting
```
z0w0, line 204: non-blocking: we should just have these all under the workspace, and cargo fmt --all --check

z0w0 (approved):

.github/workflows/ci.yml
martyportier: Agreed that a single workspace plus `cargo fmt --all --check` would simplify this. These crates are intentionally standalone workspace roots today, with separate lockfiles, to preserve their dependency-resolution boundaries; the shell workspace explicitly excludes them, so `cargo fmt --all --check` from `src-tauri` would not currently reach them. Consolidating the workspaces would be a broader dependency and CI change, so I would keep that as a separate follow-up.

PR #451, opened 2026-09-08
feat(desktop): integrate Sessions into the main window

martyportier:
## Stack

1. [PR #449 — main-window foundation](https://github.com/antiburn/antiburn/pull/449)
2. [PR #450 — collection/detail navigation](https://github.com/antiburn/antiburn/pull/450)
3. **PR #451 — Sessions integration** (this PR, targets `feat/desktop-pr2-navigation`)

## Summary

- Migrate the session list and selected-session detail into the main window.
- Reuse shared session list and detail presentation components.
- Add visibility-aware loading, stale-response protection, automatic selection, related-session navigation, and empty states.
- Suspend presentation work while the main window or Sessions section is inactive.
- Make menu-bar session cards open or focus the main window at the exact selected session.
- Preserve native and WSL session identity during deep linking.
- Remove session-detail navigation, analysis loading, and resizing state from the popover.
- Preserve privacy-safe activity and session-detail analytics.

## Scope

A dedicated design and visual-polish pass still needs to take place before this stack is ready to merge.

## Platform behavior

The deep-link path uses the shared Tauri show, unminimize, and focus behavior on macOS, Windows, and Linux. No platform-specific routing was introduced.

Pinned popovers remain on the activity list. Unpinned popovers follow the existing native focus-loss dismissal behavior.

## Validation

- Prettier, ESLint, and TypeScript passed.
- Frontend suite: 1,258 tests passed.
- Rust shell: all 1,008 tests passed.
- Main-window crate: 10 tests passed.
- Production frontend build passed.
- Packaged macOS debug app built and launched successfully.
- Windows/Linux runtime smoke testing remains required on those platforms.


apps/desktop/src-tauri/src/app_commands.rs
```diff
@@ -43,6 +43,9 @@ macro_rules! with_app_commands {
             commands::list_repositories => "list_repositories",
             commands::list_scan_roots => "list_scan_roots",
             commands::main_window_ready => "main_window_ready",
+            main_window::get_main_window_visible => "get_main_window_visible",
```
z0w0, line 46: non-blocking: why not just use built-in Tauri APIs here?

z0w0 (approved):

apps/desktop/src-tauri/src/app_commands.rs
martyportier: This command exposes a composed “can present work” state rather than the raw Tauri visibility flag. Internally it uses `is_visible()` and `is_minimized()`, and also accounts for the application being hidden on macOS. Using renderer `isVisible()` alone would leave minimized or macOS-hidden windows active, which would keep session loading and exposure tracking running. The command is also restricted to the `main` renderer.

PR #455, opened 2026-09-08
feat: add safe burn check remediation backend

z0w0:
## Summary

Burn Checks can now identify the exact local cause of a finding and prepare a safe fix action. This pull request ships the backend contract only. It does not add the remediation controls to the interface yet.

- Source-aware evidence keeps provider routes, model settings, reasoning controls, speed modes, resource use, and session relationships attached to the request that produced them.
- Exact target rows are derived before aggregate check counts. Missing or partial evidence stays unavailable instead of producing a false clean result.
- Auto Fix safely replaces reviewed old-model settings in existing Claude Code and Codex global or project configuration.
- Copy Prompt Fix creates a bounded prompt for every supported target and starts or reuses a verification watch for the correct global, project, or session scope.
- Later session evidence verifies fixes, detects recurrence, and calculates cumulative observed API-equivalent savings for old-model replacement.
- Three stable Tauri commands expose opaque target IDs to the popover without sending private target keys to the webview.

## Safety And Privacy

Auto Fix checks that the finding and source evidence are still current before it writes. It rejects ambiguous settings, changed files, unsafe file types, unsupported routes, and unreviewed replacements. Successful writes use atomic replacement, preserve exact permissions, read the complete result back, and retain a recovery state after an interrupted write.

Prompts are deterministic and limited to 8 KiB. They exclude transcript text, configuration bodies, file paths, secrets, and private selectors. Durable remediation records store bounded reviewed metadata and hashed target attribution rather than source content.

## Scope

This phase supports automatic old-model replacement for Claude Code and Codex. Prompt-based fixes cover every supported check target. The interface wiring, reasoning and resource Auto Fix actions, and wider automatic agent support remain deferred to Phase 3.

The source and check coverage baselines are recorded in `docs/session-coverage.md` and `docs/check-coverage.md`. Contract tests require every source format to remain represented.

## Validation

- Engine formatting and strict Clippy passed.
- Engine tests passed: 1,366 unit tests plus all integration and documentation tests.
- Desktop Rust formatting and strict Clippy passed.
- Desktop Rust tests passed: 1,107 default-feature tests and 1,183 analytics-feature tests.
- Frontend lint and type checks passed.
- Frontend tests passed: 1,294 tests.
- The production frontend build passed.
- Coverage contracts, repository quality checks, secret scanning, and `aislop` at 100/100 passed.


apps/desktop/src-tauri/src/store/mod.rs
```diff
@@ -1933,8 +1968,11 @@ impl Store {
     /// transcript's. Idempotent: requeuing a session already `pending` (or
     /// already claimed) just clears its retry state again.
     pub fn requeue_session_evidence(&self, key: &SessionKey) -> Result<()> {
-        let connection = self.lock();
-        mark_evidence_pending_in(&connection, key)
+        let mut connection = self.lock();
```
codesoda, line 1971: Not a blocker, but this file is getting too big now, we should break it down in next issue/s

apps/desktop/src-tauri/src/dto.rs
```diff
@@ -12,7 +12,7 @@
 
 use antiburn_local::analysis::{
     ActiveSessionsSummary, EfficiencyTotals, EvidenceValue, FAST_SPEED_KEY, ModelRun,
-    QuotaLimitKind, RepeatedContextAccounting, SessionCost, SessionEvidence,
+    QuotaLimitKind, RepeatedContextAccounting, SessionCost, SessionEvidence, SourceFormat,
```
codesoda, line 15: This file is well overdue to br broken down into submodules

AGENTS.md
```diff
@@ -44,6 +44,29 @@ Write code comments in ASD-STE100 Simplified Technical English:
 - Keep identifiers and API names unchanged.
 - Add a comment only when it states important information the code cannot show.
 
+## Session and check coverage
+
+Keep `docs/session-coverage.md` and `docs/check-coverage.md` current.
```
codesoda, line 49: We should put an AGENTS.md in the root of the folder/s that contains the coverage and checks code, calling out that touching these files require updates to the docs/... files

codesoda (approved):
NOTE: We will need to break up a few of these modules shortly

Issue #444, opened 2026-09-08
Show the last known usage dimmed while a credential waits for refresh


codesoda:
## Problem

When every credential carrier for a provider holds an expired access token, live usage reports an `Authentication` error and the row blanks after the 10-minute grace (`LIVE_USAGE_GRACE_MS`). That conflates two very different states:

- **Recoverable-expired**: a refresh token is present, no re-login is needed — the token refreshes the next time the owning tool runs.
- **Signed out / terminal**: no credentials, tombstoned entry, or a rejected refresh token — the user must log in again.

This distinction matters most where we have no CLI to delegate a refresh to. The recoverable-expired state is the **default fallback** for:

- **Pi-only machines** (Pi entries in `~/.pi/agent/auth.json` are read-only secondary carriers; recovery happens when the user next uses Pi)
- **Claude desktop app** users without the `claude` CLI
- **Codex desktop app** users without the `codex` CLI
- **Antigravity** (no refresh-only CLI lever; in-memory redeem plus the local language-server probe already cover the live path)

## Proposal (UI + error taxonomy only — no refresh behavior changes)

1. New error category (e.g. `ProviderUsageError::CredentialExpired`) emitted when carriers exist, all access tokens are expired, and at least one carrier has a refresh token present. Presence-only check — we never read or use the refresh token value.
   - `anthropic_fetch.rs`: `fetch_from_carriers` currently collapses this into `Authentication` (the `!live` branch).
   - `codex_fetch.rs`: the expired Pi entry / expired auth.json cases.
2. Frontend: when the error is `CredentialExpired`, keep showing the last-known snapshot **dimmed** past the 10-minute grace instead of blanking, with a tooltip like: "Session expired — showing usage from 2 hours ago. Refreshes next time you use Claude." Tailor the tool name to the carriers present (Pi vs CLI vs unknown).
3. Distinct terminal state: carriers absent or refresh rejected → actionable "sign in again" presentation, not the dimmed state.
4. Respect design tokens per `apps/desktop/design.md` (no hard-coded colors/durations for the dimmed treatment).

## Notes

- This is a prerequisite/companion for the Codex and Claude CLI-delegated-refresh issues: their "binary absent" path falls through to this state.
- The dimmed snapshot can be many hours old (pre-sleep); the tooltip must carry the captured-at time to stay honest.
- Analytics: add an event for entering/leaving the recoverable-expired state per provider so we can measure how often users sit in it (see docs/analytics.md contract).

PR #458, opened 2026-09-08
feat(desktop): delegate expired Claude token refresh to the Claude CLI

### LINKED ISSUE #443: Claude live usage: delegate expired-token refresh to the Claude CLI
## Problem

`provider_usage/live/sources/anthropic_fetch.rs` deliberately never refreshes OAuth tokens (Anthropic rotates refresh tokens on redemption, and we do not write the Keychain or `~/.claude/.credentials.json`). Correct — but it means that after the access token expires (routinely, e.g. overnight sleep), `fetch_from_carriers` reports `Authentication` with no network call and Claude usage disappears until the user happens to run Claude again.

## Proposal

Same pattern as the Codex companion issue: the CLI owns the credential, so make the CLI do the refresh.

1. **Token still valid** → use it directly for `GET api.anthropic.com/api/oauth/usage`. Done.
2. **Token expired (refresh token present)** → launch the `claude` CLI in a PTY and issue a lightweight command (e.g. `/status`) purely to make it refresh and persist its own credential — no model request, no token spend.
3. **Verify the refresh happened** by fingerprinting the credential carrier (Keychain item / `.credentials.json` content) before and after the spawn and confirming it changed, then re-read and retry the usage call once.

## Guardrails

- Only expired/401 credential states trigger the touch; network/5xx failures do not.
- Probe that the `claude` binary exists first; if absent (e.g. Pi-only or Claude-desktop-only machines), fall through to the recoverable-expired UI state (companion UI issue).
- On macOS, skip background-initiated touches that could raise a Keychain access prompt; allow them only on a user-initiated refresh (popover open / manual refresh).
- Cooldown the touch (minutes-scale) with in-flight dedup so one expired credential cannot spawn PTYs repeatedly.
- Classify `invalid_grant`-style terminal failures as "run `claude` to log in again", blocked until the credential material changes on disk, distinct from transient backoff.

## Out of scope

- Pi's `~/.pi/agent/auth.json` entry stays a read-only secondary carrier; it recovers when the user next uses Pi.

codesoda:
Closes #443

## What

When every Claude credential carrier has expired, a **user-initiated** refresh now delegates the refresh to the CLI that owns the credential: spawn `claude` in a PTY, type `/status` (token refresh without a model request), verify the credential carrier changed and settled, then re-read once and retry the usage call once.

New module: `provider_usage/live/sources/claude_touch.rs`. Integration lives in `anthropic_fetch.rs`; shared-file changes (`sources/mod.rs`, `cooldown.rs`) are minimal and additive. `pi_auth.rs`, `codex_fetch.rs`, and `codex_app_server.rs` are untouched.

## Verification protocol (both issue comments)

- Fingerprint the carrier **before** the spawn, poll **after** it until quiescent: changed from pre-spawn *and* re-observed unchanged for 4 consecutive polls (~2 s). The first change is never accepted.
- 500 ms poll floor, 30-poll hard cap (~15 s deadline); the PTY child is killed when verification concludes, either way.
- **Metadata only, never the secret**: on macOS the poll runs `security find-generic-password` *without* `-w` (attributes only, never hits the item ACL — cannot prompt, by construction). The `.credentials.json` fallback is fingerprinted by content hash. The secret is read exactly once, after the change has settled.

## Guardrails

- Triggers only on expired/401 credential states — never on network/5xx failures.
- `claude` binary probed first; absent (Pi-only / Claude-desktop-only machines) keeps today's `Authentication` behavior (the recoverable-expired UI is #444).
- **Prompt policy**: background polls never reach the touch path; the caller's `max_age` already distinguishes the popover (50 s) from the background monitor (5 min), so no new plumbing was needed.
- Minutes-scale touch cooldown (5 min) with in-flight dedup, modeled on `cooldown.rs`'s documented pattern.
- `invalid_grant`-style terminal failures (refresh settled but the credential is still dead, or the retried call still 401s) block further touches until the credential material changes on disk — distinct from transient backoff.
- Keychain secret cached until its own `expiresAt`; a live token's secret is never re-read.

## Dependency

Adds `portable-pty` — nothing in the tree can allocate a PTY, and the CLI refreshes its token only when it runs interactively. Rationale in the module doc; `THIRD_PARTY_NOTICES` regenerated; `cargo deny check bans licenses advisories` clean.

## Tests

- `claude_touch`: settled-change-is-not-the-first-change, revert-never-settles, timeout-at-poll-cap, binary-absent, unobservable-carrier, gate dedup/cooldown, terminal-until-material-changes, poll floor/cap bounds.
- `anthropic_fetch`: expired→touch→settled→secret-read-once→retry-once; never-settles stays `Authentication`; binary-absent; background poll never touches (panicking env); terminal gate blocks re-spawn then recovers after re-login without a touch; Pi carrier alone never triggers a touch; live cached Keychain secret not re-read; user-initiated ceiling splits popover from monitor.

`cargo fmt`, `cargo clippy --all-targets` (with and without `analytics`) clean; `cargo test --lib` 1027 passed.


apps/desktop/src-tauri/src/provider_usage/live/sources/claude_touch.rs
```diff
@@ -0,0 +1,680 @@
...
+//! carrier has expired, this module makes *the CLI* refresh and persist its
+//! own credential: it spawns `claude` in a PTY, types the lightweight
+//! `/status` command — which triggers a token refresh without a model
+//! request — and tears the process down again. No token of the CLI's is ever
+//! redeemed, written, or held here.
+//!
+//! # Why a PTY, and why this crate
+//!
+//! The CLI refreshes its credential when it runs interactively; piped stdio
+//! puts it in a non-interactive mode that does not. Nothing already in the
```
daveslutzkin, line 18: seems weird that in non-interactive mode it would just fail to refresh the creds and so just fail to work?

daveslutzkin (approved):

songcarver:
So far, no issues, usage meter showing as expected. It didn't ask me for access to keychain (not sure if it was going to)

PR #488, opened 2026-09-10
feat: Burn Check main window, with remediation

z0w0:
## Summary
This PR completes the Burn Check remediation flow. It turns detected token-waste patterns into clear, local-only actions that help people understand the problem, make a fix, and confirm the result.

### What people can do
- Open any failed Burn Check and see a short explanation of the waste pattern and its likely impact.
- Review affected configuration items when a check has named targets, such as unused skills or MCP servers.
- See a small set of related session examples with their real local titles.
- Copy a focused fix prompt for one item, or one complete prompt for every item in a failed check.
- Apply supported configuration fixes directly from the app, then see applied, verified, or still-needs-attention states.
- See estimated token savings only when the available evidence supports an estimate.

### How remediation works
- Groups related findings into useful fixes instead of repeating the same advice for each session.
- Covers all Burn Check categories, including old models, excessive thinking, cache misses, unused skills and MCP servers, routes, deep agent trees, and repeated context.
- Falls back to a safe check-level prompt when a finding no longer has a current configuration target.
- Watches changed configuration files and rechecks affected items after a fix.
- Does not mark an unsupported or resource-only finding as fixed, and does not claim savings for it.

### Privacy and safety
- The UI and analytics use derived remediation data only.
- Raw session IDs, paths, selectors, configuration bytes, prompts, and session titles do not enter analytics or public report payloads.
- Local source paths appear only in a prompt the person explicitly copies.
- Automatic fixes are limited to supported, exact configuration changes.

## Validation
- cargo test --test check_coverage_contract
- node scripts/check-design-drift.mjs
- aislop scan --changes --base origin/main
- Earlier feature validation: desktop lint, type check, tests, build; desktop and engine Rust fmt, clippy, and tests; secrets check.

## Screenshot

Screenshot 2026-09-10 at 8 56 54 pm


apps/desktop/src-tauri/src/agent_config/vendors/claude.rs
```diff
@@ -0,0 +1,402 @@
...
+                        scope: ConfigScope::Project,
+                        operation,
+                    });
+                }
+            }
+        }
+
+        let path = home.join(".claude/settings.json");
+        if !path_entry_exists(&path)? {
+            return Err(ConfigUnavailableReason::MissingConfig);
```
codesoda, line 53: So we're only "adding" to an existing config, not creating a config for them?

apps/desktop/src-tauri/src/agent_config/vendors/codex.rs
```diff
@@ -0,0 +1,369 @@
...
+                    path,
+                    safety_root: root.to_owned(),
+                    scope: ConfigScope::Project,
+                    operation,
+                });
+            }
+        }
+
+        if !path_entry_exists(&global)? {
+            return Err(ConfigUnavailableReason::MissingConfig);
```
codesoda, line 52: Same question as claude....

apps/desktop/src-tauri/src/agent_config/vendors/claude.rs
z0w0: I intend to expand on this in future PRs (this one's already too big)

apps/desktop/src-tauri/src/agent_config/vendors/codex.rs
z0w0: I intend to expand on this in future PRs (this one's already too big)

apps/desktop/src-tauri/src/remediation/target.rs
```diff
@@ -0,0 +1,431 @@
...
+    secret: &[u8; 32],
+    finding: &CurrentFinding,
+    canonical_workspace: Option<&Path>,
+) -> TargetIdentity {
+    let agent = crate::agents::kind_from_slug(&finding.agent)
+        .expect("the caller validates the finding agent");
+    let workspace_key = canonical_workspace
+        .and_then(Path::to_str)
+        .map(|workspace| hashed_parts_with_secret(secret, b"workspace", &[workspace]));
+    let (mut scope_kind, mut scope_key) = if let Some(workspace_key) = workspace_key.as_ref() {
```
codesoda: **P1 — Preserve session/worker scope for non-durable findings.** When `canonical_workspace` exists, this unconditionally sets every finding to `project` scope. For `SessionsOverDepth`, `OverpoweredSubagents`, and `CacheChurn`, `Finding::canonical_identity` does not include session or worker identity, so equal findings from separate sessions share the `group_key` in `remediation/mod.rs:211-216`. That merges occurrences and quantities and combines representative paths in prompts. `docs/remediation.md:93-95` says session and worker targets stay scoped when no safe durable control exists. Please preserve session/worker scope for these detectors, or include their durable identity.

apps/desktop/src/views/main-window/burn-checks/BurnCheckDetail.tsx
```diff
@@ -0,0 +1,315 @@
...
+  const copy = async () => {
+    if (busy || copied) return
+    const startedKey = currentKey
+    setBusy(true)
+    setStatus(null)
+    let nextPrompt = prompt
+    try {
+      if (!nextPrompt) {
+        const outcome =
+          promptTargets.length === 0
```
codesoda: **P1 — Do not invoke the check-level fallback when exact targets exist.** `promptTargets.length === 0` is true both when `targets` is empty and when every exact target's prompt is unavailable. However, `copy_prompt_fix_burn_check` rejects the request when current findings exist (`apps/desktop/src-tauri/src/remediation/mod.rs:366-377`), and the target list is built from those same findings. The button therefore always returns unavailable and refreshes into the same state. Gate the fallback on `targets.length === 0`; otherwise show the unavailable exact-target state.

crates/antiburn-local/tests/check_coverage_contract.rs
```diff
@@ -1,249 +0,0 @@
-use std::collections::BTreeMap;
```
codesoda, line 1: **P2 — Restore the coverage contract test.** This deletes the exhaustive contract test that covered partial-evidence clean gating, uncharacterized sources, badges, and direct findings across all source formats. I could not find an equivalent replacement in this PR. The stated validation command also now fails with `no test target named check_coverage_contract`. Please restore or relocate the contract test and update the validation command if needed.

codesoda (commented):
Review findings requested by the author. All six findings are actionable; the duplication findings reference both duplicated implementations.

apps/desktop/src/views/main-window/burn-checks/BurnChecksReport.tsx
```diff
@@ -0,0 +1,278 @@
...
+import { Skeleton } from "../../../components/ui/Skeleton"
+import type { ChecksCategoryPayload, ChecksReportPayload } from "../../../lib/insightsIpc"
+import { checksHeroPresentation, checksPresentation } from "../../../lib/presentation/checks"
+import { checkRowPresentation } from "../../checks/checkUi"
+import type { BurnChecksSession, BurnChecksSnapshot } from "../BurnChecksSession"
+import { BurnCheckDetail } from "./BurnCheckDetail"
+import { BurnCheckTargetDetail } from "./BurnCheckTargetDetail"
+import { BurnChecksSavings } from "./BurnChecksSavings"
+
+function DisclosureChevron({ open }: { open: boolean }) {
```
codesoda: **P2 — Reuse the existing `DisclosureChevron` instead of duplicating it.** This local implementation duplicates the exported `DisclosureChevron` in `apps/desktop/src/views/main-window/burn-checks/BurnCheckTargetPresentation.tsx:9-21`; `BurnChecksSavings.tsx` already imports that shared symbol. Please import and reuse the existing component so the two implementations cannot diverge.

apps/desktop/src/views/main-window/burn-checks/BurnCheckDetail.tsx
```diff
@@ -0,0 +1,315 @@
...
+  overpoweredSubagents: "Some helper work used more model power than it needed.",
+  unusedMcpServers: "Some MCP servers were loaded but not used.",
+  unusedBuiltInTools: "Some built-in tools were loaded but not used.",
+  unusedSkills: "Some skills were loaded but not used.",
+  oldModelUsage: "Some sessions used an older model when a newer one was available.",
+  overuseOfFastMode: "Some work paid for speed it did not need.",
+  cacheChurn: "Some sessions kept paying to reload the same context.",
+}
+
+function Samples({ targets }: { targets: BurnCheckTargetPayload[] }) {
```
codesoda, line 26: **P2 — Consolidate the duplicate sample-session interaction logic.** `Samples` here duplicates the state, disclosure, busy navigation, error handling, and `SessionSampleRow` rendering in `apps/desktop/src/views/main-window/burn-checks/BurnCheckTargetPresentation.tsx:77-137` (`SampleSessions`). The multi-target deduplication is unique here, so extract a shared lower-level component and pass the normalized sample list instead of maintaining both implementations.

apps/desktop/src-tauri/src/remediation/display.rs
```diff
@@ -0,0 +1,409 @@
...
+        })
+        .collect()
+}
+
+pub(super) fn burn_check_display_facts(target: &CachedTarget) -> BurnCheckDisplayFacts {
+    use antiburn_local::remediation::BuiltInToolTokens;
+
+    let finding = &target.findings[0].finding;
+    let observed_at_ms = target.findings[0].observed_at_ms;
+    let (resource_kind, resource_identity, current_value, replacement_value, mut quantity, unit) =
```
codesoda: **P2 — Centralize the duplicate quantity/unit mapping.** The `FindingCause` match here repeats the `FindingCause -> (quantity, BurnCheckQuantityUnit)` mapping implemented by `finding_quantity` at `apps/desktop/src-tauri/src/remediation/display.rs:336-366`. This match also computes the other display fields, but its quantity/unit arms should call the helper; otherwise future detector or semantic changes can make single-finding and aggregated displays disagree.

apps/desktop/src/views/main-window/burn-checks/BurnCheckDetail.tsx
z0w0: Fixed in c540ca58. The check fallback now runs only when the target list is empty. If exact targets exist but none have a prompt, the check action is hidden. Added a frontend regression test.

apps/desktop/src-tauri/src/remediation/target.rs
z0w0: Fixed in c540ca58. Depth and cache findings now keep session scope. Subagent findings use a worker scope keyed by the session and worker identity. Added regression coverage for separate sessions.

apps/desktop/src/views/main-window/burn-checks/BurnChecksReport.tsx
z0w0: Fixed in c540ca58. `BurnChecksReport` now imports and uses the shared `DisclosureChevron`.

crates/antiburn-local/tests/check_coverage_contract.rs
z0w0: Restored the full check coverage contract in c540ca58. `cargo test --test check_coverage_contract` passes again.

apps/desktop/src-tauri/src/remediation/display.rs
z0w0: Fixed in c540ca58. `burn_check_display_facts` now calls `finding_quantity` for the quantity and unit mapping.

apps/desktop/src/views/main-window/burn-checks/BurnCheckDetail.tsx
z0w0: Fixed in c540ca58. Both surfaces now use `SampleSessions`; the check detail only handles its target sample deduplication.

codesoda (approved):

PR #516, opened 2026-09-13
Add priced unused-resource checks to the session Hygiene tab

daveslutzkin:
Stacked on #514. PR 2 of `docs/plans/prefix-cost-estimates.md`: the per-session Hygiene tab now carries the three prefix checks (unused MCP server, unused built-in tool, unused skill) with a dollar figure per named resource.

## What changes

- `BadgeId` grows from six to nine. The three new badges map to the existing `UnusedMcpServers`, `UnusedBuiltInTools` and `UnusedSkills` detectors, so the finding rule is the detectors' own.
- The session Hygiene payload gains one evidence arm per badge, each a list of `{ name, costUsd }`. The cost is the resource's token count times every observed model's assistant request count times that model's live cache-read rate, summed over `models.by_model`. Sub-agent requests are included because the by-model query is not scoped to the main context. `costUsd` is null when no observed model resolves in the pricing table; the name still reports.
- Built-in tools apply the detector's own filter: real, non-deferred, never-invoked, non-situational definitions.
- The Hygiene tab renders `name — ~$X.XX` per resource as the finding's detail lines.

## Presentation decision worth a look

The engine policy says no reader proves a full historical resource inventory, so these three detectors never read Clean (`clean_facts_complete`). Left alone, every session would show three permanent "not assessed" marks, the session list could never reach "All Burn Checks passed", and the reason copy ("couldn't read the whole session log") would be false.

This PR presents the three checks only when they are a finding. The alternative, letting M/B/K read Clean per session when the inventory is complete, would change the coverage contract and the aggregate report's clean counts. That is a maintainer decision and is recorded in the plan's deviations.

## Verification

Engine crate: fmt, clippy `-D warnings`, full tests. Shell crate: same, with the seven known local-only failures. Frontend: type-check, lint, vitest, knip, prettier, design-drift. `check_coverage_contract` passes.

https://claude.ai/code/session_018n3uiYEqE7sfpDGWnR6Vip


daveslutzkin:
Closing. Per-session "unused" is a fact about one session, not evidence for a config change, and these checks can never read clean. The aggregate Burn Checks report (#514) already carries the recommendation and the summed cost. The per-session dollar figure will return as an informational line on the Cost tab in a smaller PR, reusing the pricing helper from this branch.

PR #547, opened 2026-09-15
fix(deps): upgrade rustls to 0.23.45

codesoda:
## Summary
- Upgrade the desktop lockfile rustls entry from 0.23.44 to the patched 0.23.45 release.
- Refresh the generated third-party notices for rustls.

This is a stacked dependency-only PR for #541. It does not modify the existing PR branch.

## Validation
- cargo-deny bans/licenses and advisories
- cargo fmt --check (desktop workspace and companion crates)
- Clippy with warnings denied, debug/release, including analytics
- Rust tests, debug/release, including analytics
- pnpm install --frozen-lockfile, secrets, slop:all, notices, design drift, and diff checks


codesoda:
Superseded by #548, which contains the same dependency-only change on a branch based directly on main.

PR #570, opened 2026-09-16
Learn a limit factor for model-scoped weekly windows

daveslutzkin:
## Summary

Phase 2a of the quota usage work. The limit-factor learner now tracks a
lane for a provider's supplemental weekly window scoped to one model —
Anthropic's "Fable" limit today — alongside the existing five-hour and
weekly account-wide lanes.

- **Lane identity.** `lane_for_period` replaces `lane_for_window_role`,
  deriving a lane from a period's role, kind, id, and scope instead of
  role alone: `primaryShort`/`primaryLong` with `scope_key = 'account'`
  keep the two fixed lanes; a `supplemental` weekly window whose
  `scope_key` starts `model:` maps to its own `model:<slug>` lane, keyed
  by the slug in the window id (falling back to slugifying the scope
  label when the window id has another shape).
- **Candidate periods and observations.** `provider_limit_candidate_periods`
  now admits a supplemental model-scoped weekly period alongside the two
  account-wide primary lanes. `latest_observation_plan` reads a model
  lane's plan from its own supplemental window's observation, not the
  account window's — the old code could only reverse a lane to a window
  role, which collides across different model-scoped windows.
- **Model matching and dollar filtering.** `model_matches_scope` slugifies
  a provider's scope label and checks it against a contiguous run of the
  turn model id's own hyphen segments (stripping a `[..]` context-window
  tag first). `attributed_turn_dollars_between` gains a `model_scope`
  parameter that filters priced turns to the matching model in Rust,
  after the query, so the pricing path stays identical. The factor
  learner passes a model lane's scope label through
  `build_period_samples` and `compute_residual`, so a model lane's
  samples and residual price only that model while the account-wide
  weekly lane keeps pricing every model.
- **Schema.** `user_version` 49 → 50 rebuilds
  `provider_limit_factor_sample` and `provider_limit_factor_point` so
  their `lane` `CHECK` admits `model:<slug>`, since SQLite cannot alter a
  `CHECK` constraint in place. Every column, the `UNIQUE` constraint, and
  both lane indexes carry over unchanged.
- **Diagnostics.** `latest_residuals_by_lane` used to resolve a residual's
  lane from its period's window role alone via a SQL self-join; two
  different model-scoped windows can share `window_role = 'supplemental'`,
  so it now reads every field `lane_for_period` needs and resolves the
  lane, keeping the newest reading, in Rust.

## Codex supplemental-window scope finding

Codex's `additional_rate_limits` entries (`provider_usage/live/codex.rs`,
`supplemental_window`) always set `scope: UsageScope::Model(limit_name)` —
there is no `UsageScope::Account` branch for a Codex supplemental window.
So every Codex supplemental window qualifies for a lane under the exact
same rule as Anthropic's, with no provider special-casing: a Codex
feature-scoped window whose stated duration is exactly 604,800 seconds
(weekly) gets a `model:<slug>` lane; a non-weekly (rolling) Codex
supplemental window carries no lane, the same restriction Anthropic's
weekly-only rule already applies.

## Analytics decision

`antiburn.limit_factor_observed`'s `detail` property is free text (not a
typed enum), so per the spec every `model:<slug>` lane collapses to the
single closed value `model` — never the model name or its slug. Updated
the catalog entry in `docs/analytics.md` and added a dated review note to
`docs/analytics-measurement.md`.

## Deviation from spec

Spec item 12 asked to fix `docs/plans/limit-factor-estimation.md`'s front
matter and add a dated note about model-scoped lanes. That file does not
exist on this branch — it was removed by the "remove planning artifacts"
chore (#560) before this branch was cut from `feat/quota-retention-history`.
Per the coordinator's instruction mid-task, this item is skipped entirely.

## Tests

- `store::provider_limit::tests`: `lane_for_period` (all four branches:
  five-hour, weekly, model-scoped via window id, model-scoped via
  scope-label fallback), candidate periods admitting a model-scoped
  weekly period and excluding an account-scoped supplemental period,
  `latest_observation_plan` reading a model lane's own supplemental
  reading rather than the account window's, and a per-step v50 migration
  test (existing sample/point rows survive the rebuild, `model:fable`
  now inserts, `bogus` still fails the check — kept for review, may be
  deleted after merge per repo convention; the ladder test is durable).
- `provider_usage::factor::tests`: `model_matches_scope` against every
  example in the spec, and a new test inserting a Fable turn and an Opus
  turn inside a delta pair on a `model:Fable` period — the Fable lane's
  sample prices only the Fable turn, the account-wide weekly lane's
  sample on a matching `primaryLong` period prices both.
- `analytics::enabled::tests`: a model-scoped lane reports the single
  `detail` value `model`, never the slug.

## Check status

- `cargo fmt --check` — clean
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo clippy --all-targets --features analytics -- -D warnings` — clean
- `cargo test --no-fail-fast` — 1373 passed, 7 failed (pre-existing,
  unrelated: `agent_config`, `insights_report`, `remediation` — confirmed
  identical on `feat/quota-retention-history` before this branch's changes)
- `cargo test --no-fail-fast --features analytics` (with
  `ANTIBURN_ANALYTICS_URL`/`ANTIBURN_ANALYTICS_OPERATOR` set per
  CONTRIBUTING.md) — 1478 passed, 7 failed (same pre-existing failures)
- `crates/antiburn-local`: `cargo fmt --check`, `cargo clippy --all-targets
  -- -D warnings`, `cargo test --no-fail-fast` — all clean/green, untouched
  by this change
- `pnpm run slop:all` — score 100, 0 issues
- `pnpm run secrets` — clean

## Summary by CodeRabbit

* **New Features**
  * Added weekly usage limits scoped to individual models alongside account-wide limits.
  * Model-specific attribution, observations, and diagnostics now distinguish matching model activity from other usage.
* **Bug Fixes**
  * Improved limit-factor learning and analytics categorization for model-scoped usage without exposing model names or slugs.
* **Data & Compatibility**
  * Existing provider-limit data is preserved and upgraded automatically to support model-scoped lanes.
  * Analytics reporting distinguishes model-scoped events without reclassifying historical data.


z0w0 (approved):

daveslutzkin:

PR #571, opened 2026-09-16
Add quota period, meter, and contribution queries

daveslutzkin:
## Summary

Phase 2b of the quota usage feature: the store queries and Tauri commands
the quota screen will read from, stacked on phase 2a's per-model lanes
(#570) and phase 1's retention history.

- **`provider_usage::quota::resolve_periods`**: turns a lane's observed
  `provider_usage_period` rows into a timeline of quota windows. Each
  boundary carries a `BoundarySource`: `reported` (the provider stated it),
  `derived` (computed from the other boundary and the lane's nominal
  duration), `cadence` (extrapolated by whole weeks from the most recent
  observed weekly reset, dropped where it would overlap an observed
  period by more than the 5s reset jitter), or `turnGap` (inferred from a
  gap in local turn activity at least as long as the five-hour lane's
  duration; capped at 400 inferred windows per call).
- **Bucketed contributions**: a new SQL variant of the existing per-session
  attribution query adds `t.ts_ms / 900000` to the `SELECT`/`GROUP BY`,
  reusing the same joins, fences, `LIMIT`, and pricing path
  (`lookup_turn_pricing` → `calculate_cache_write_cost` for 1-hour cache
  writes) as the existing per-session query. Unbound sessions are kept
  (reported separately) rather than dropped. **A bucket that straddles a
  period reset is assigned to the period containing the bucket's start**,
  not its end — documented on `attributed_turn_dollars_by_bucket`.
  `quota_contribution_bind_rate` is logged once per call via
  `tracing::info`, with bound/unbound/other-account counts only (no keys,
  no dollars).
- **Percent**: `usd / factor_point_at(...).usd_per_percent`, loaded once
  per call as an ordered point series and resolved per bucket by binary
  search (`slice::partition_point`) instead of a query per bucket.
- **Three commands**: `get_quota_accounts` (every observed
  `(provider, account)` and its lanes, each lane carrying its current open
  period — the lane's most recent period whose reset is after now, start
  derived the same way the resolver derives it), `get_quota_usage` (one
  lane's periods, meter readings, and per-session contributions over a
  range), `get_session_quota` (one session's estimated entries by provider
  and lane; a session with no resolved account still gets one entry,
  `confidence: "unbound"`, no period).
- **Range cap**: `get_quota_usage` rejects a range over 35 days.
- Frontend: TypeScript mirrors of the three DTOs plus IPC wrappers
  (`getQuotaAccounts`, `getQuotaUsage`, `getSessionQuota`) — types and
  wrappers only, no UI in this phase.

## Deviations from spec

- **File-size split**: adding the three commands pushed `commands.rs` past
  the `aislop` file-size cap (Rust files: 3750 effective LOC). Split it
  into `commands/mod.rs`, `commands/quota.rs` (the new commands), and
  `commands/local_usage.rs` (the pre-existing local/live-usage and
  session-limit-allocation commands that shared the same stretch of the
  file). Each `#[tauri::command]` fn is now registered in
  `app_commands.rs` by its submodule path (`commands::quota::...`,
  `commands::local_usage::...`), since `generate_handler!` needs the
  literal path the macro expanded at — a plain `pub(crate) use` re-export
  doesn't carry the macro's hidden `__cmd__`/`__tauri_command_name__`
  items across module boundaries. Not in the spec, but required to keep
  `pnpm run slop:all` green.
- **Frontend file placement**: put the new types and IPC wrappers in
  `providerUsageIpc.ts` rather than `ipc.ts` — `ipc.ts` is already at its
  own `aislop` cap, and `providerUsageIpc.ts` already mixes payload types
  with wrappers for other commands (matching the established
  `mainWindowIpc.ts` pattern). Uses `isTauri()` directly rather than
  `ipc.ts`'s `hasShell()`, to avoid a circular import.
- **Mid-task addition** (from the coordinator, applied): each
  `QuotaLanePayload`/`QuotaLane` carries `currentPeriod`/`current_period`,
  covered by a store test (`quota_accounts_reports_label_has_factor_and_the_current_open_period`)
  with one lane whose period is still open and one whose latest period
  has already reset.
- **Unbound session's lane in `get_session_quota`**: the spec doesn't say
  what lane id an unbound entry should carry (there is no resolved account
  to look up a lane against). Used `LANE_WEEKLY` as a placeholder; flagging
  as an interpretive choice for review.

Everything else follows the spec as written.

## Tests

- `provider_usage::quota`: one test per boundary source (reported,
  derived, cadence forward/backward with an observed slot suppressing a
  cadence slot, turn-gap opening two windows six hours apart and not
  opening one for a turn inside an observed period), the 400-window cap,
  empty inputs. 11/11 passing.
- `store::provider_limit`: bucketed query groups by 15-minute bucket,
  excludes a turn on an unpublished fence, drops another-account sessions
  and keeps unbound as unbound, model-scope filter, `quota_periods_for_lane`
  returns overlapping periods plus the anchor before the range,
  `factor_points_for_lane` orders by `effective_at_epoch`,
  `attributed_turn_epochs` rounds to the minute and respects the two-step
  account rule, `quota_accounts` reports label/has-factor/current period.
  8 new tests, 27/27 total in the module passing.
- `commands`: `get_quota_usage` end to end on an in-memory store (one
  period, three observations, two bound sessions, one unbound) asserting
  bucket assignment, `sessions` ordering, `unattributed`, and
  `estimated_percent`; rejects a 36-day range; `get_session_quota` for a
  session spanning a five-hour reset returns two entries; an unbound
  session reports one `"unbound"` entry. 5/5 passing.
- `dto`: camelCase field names and boundary-source/confidence wire strings
  for both new payloads. 2 new tests, 16/16 total in the module passing.

## Checks

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`: clean,
  with and without `--features analytics`.
- `cargo test --no-fail-fast` (desktop shell): **1405 passed; 0 failed; 1
  ignored**. With `--features analytics`
  (`ANTIBURN_ANALYTICS_URL`/`ANTIBURN_ANALYTICS_OPERATOR` set per
  CONTRIBUTING.md): **1510 passed; 0 failed; 3 ignored**.
- `crates/antiburn-local` (workspace root, untouched by this change):
  `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` clean;
  `cargo test --no-fail-fast`: every binary **0 failed** (1449 passed in
  the main lib suite; the rest split across smaller integration binaries).
- `pnpm --filter @antiburn/desktop lint`, `type-check`: clean.
- `pnpm --filter @antiburn/desktop test`: **130 files, 1726 tests passed**.
- `pnpm run slop:all`: score 100, 0 errors (after the file-size split
  above).
- `pnpm run secrets`: clean.

## Timing

Copied the real local database (`~/Library/Application Support/ai.antiburn.desktop/antiburn.sqlite3`,
686 MB, 356,865 `turn` rows spanning the full history) to a scratch
directory and ran the new bucketed query
(`ATTRIBUTED_TURN_BUCKET_SQL`) directly in `sqlite3` with `.timer on`,
provider `anthropic`, `LIMIT 20001`:

| Range | Wall time |
| --- | --- |
| 7 days | ~0.11s (0.106–0.111s over 3 runs) |
| 30 days | ~0.73s (0.730–0.746s over 3 runs) |

`EXPLAIN QUERY PLAN` confirms the range scan uses the `turn_usage_timestamp`
index, every join is indexed (including the correlated
`session_provider_account` subquery), and the only non-indexed step is
the temp B-tree for the final `GROUP BY`.


z0w0 (approved):

daveslutzkin:

PR #572, opened 2026-09-16
Add the Quota screen

daveslutzkin:
## Summary

Adds a new "Quota" section to the main window, between Overview and
Checks, showing one provider account/lane's quota windows as a burnup
chart plus a "Top sessions" list.

- **What it shows**: an account and lane picker (weekly/five-hour/model
  lanes), a range preset control, a summary strip (meter peak,
  estimated total, gap or over-estimate, last reading age), the burnup
  chart, and a "Top sessions" list with an "Unattributed" row when
  applicable. Clicking a session opens it in Sessions via
  `activitySession.openRelated`.
- **Range preset rule**: `thisWeek` uses the lane's current period when
  one is open, `lastWeek` shifts that window back seven days, and
  `last30Days` is capped at 35 days of span. The five-hour lane's
  `thisWeek`/`lastWeek` borrow the account's weekly lane's current
  period as their reference window, since the five-hour lane itself
  has no week-shaped period.
- **Bucket and reset rendering rule**: each period renders as a
  stacked-area burnup (top 5 sessions + "other" + "unattributed") plus
  a provider-meter line, both reset to zero at the period's boundary.
  Session, "other", and unattributed columns all accumulate from their
  own per-bucket data (`contributions` and `unattributedBuckets`), so
  every stacked layer is a real time series, not an approximation.
  The meter line interpolates linearly between two authoritative
  samples at most three hours apart, and stays null before the first
  sample, after the last, across a wider gap, and across a period
  boundary. Rows sit on a 15-minute grid, plus any bucket or sample
  timestamps a period actually reports, so a burst of activity does
  not get smoothed away.
- **Analytics**: adds a `quota` surface to `surface_viewed` and
  `surface_state_observed`, both frontend (`Surface` union in
  `ipc.ts`) and Rust (`Surface`/`StateSurface` in
  `analytics/event.rs`), documented in `docs/analytics.md` and
  `docs/analytics-measurement.md`. No provider, account, lane, dollar
  figure, or session identity is collected; the exposure identity
  string used for local de-duplication never leaves the renderer.
- Chart primitives (`ChartKey`, `chartLabels`, `useChartResize`) are
  cherry-picked unmodified from `feat/cost-burnup-buckets`, in their
  own first commit, since this branch predates that PR merging.
- Rebased onto `feat/quota-periods` at `a2aad301`, which added
  `QuotaPeriodPayload.unattributedBuckets`; the unattributed column
  now accumulates from that instead of the ramp an earlier revision
  of this PR used.

## Deviations from the phase 3 spec

- `rangeForPreset(preset, lane, now)` gained an optional fourth
  parameter, `weeklyLane`, so the five-hour lane's `thisWeek`/
  `lastWeek` can borrow the account's weekly lane's `currentPeriod`.
  The spec's three-argument signature has no way to pass that lane in.
- `QuotaBurnupChart` renders the `ChartKey` legend inside itself,
  rather than the caller rendering the key separately. The spec's
  chart props explicitly include `onHighlight`/`onPin`, which only
  make sense if the key lives inside the chart component.
- The Quota tab's selection is local `useState` in `MainWindowView`,
  not a new value in the shared `MainWindowSectionId`/
  `MainWindowSection` type used for cross-window navigation IPC. The
  spec gives Quota no cross-window entry point, and widening that
  shared contract risked IPC drift for no benefit. A
  render-time "adjust state" comparison (a `useState`-held previous
  value, not a ref or effect) clears the local selection whenever a
  genuine cross-window navigation event arrives, and a separate
  `quotaVisited` flag keeps the section mounted once visited, like
  every other section.
- Screenshots are omitted: this environment has no way to run the
  Tauri dev app and capture its window.

## Tests

- `quotaSeries.test.ts` — 23 tests (range presets, series building,
  meter interpolation and its gap cutoff, per-bucket unattributed
  accumulation, top-session merging, unattributed totals,
  latest-period/sample helpers).
- `QuotaSession.test.ts` — 9 tests (initial load, lane/account
  switching, coalesced live-usage refresh, scan-triggered refresh,
  error handling).
- `QuotaBurnupChart.test.tsx` — 6 tests (stacked areas, reset line
  styling, now-line, inferred-reset caption, no-factor fallback,
  highlight dimming).
- `QuotaView.test.tsx` — 8 tests (loading/empty/error states, range
  control reload, top-session row click, unattributed row
  visibility).
- `MainWindowView.test.tsx` — updated for the new tab count, plus new
  tests covering Quota's local-only selection, its clearing on a
  cross-window navigation request, and staying mounted after
  navigating away.
- Rust: `Surface`/`StateSurface` compile under both feature
  configurations; existing `commands::quota`, `provider_usage::quota`,
  and `store::provider_limit` suites (25 tests) still pass unchanged.

Full frontend suite: 1774/1774 passing. Rust `cargo test --no-fail-fast`
passes both with and without `--features analytics` (a pre-existing,
environment-gated failure in `analytics::enabled::*` reproduces
identically on `feat/quota-periods` before this branch's changes, and
is unrelated to this work).


z0w0 (approved):

daveslutzkin:

PR #584, opened 2026-09-18
fix(desktop): let Wayland title-bar buttons receive clicks

martyportier:
## Problem

On a native Wayland run, the title-bar Close and Minimize buttons of the main, Settings, and setup windows do nothing. The settings window is non-resizable, so the known "double-click to unstick the buttons" escape hatch does not even exist for it.

## Cause

tao 0.35 draws its own client-side title bar on Wayland: a GTK `HeaderBar` inside a `GtkEventBox` with the `above-child` flag set ([tao#899](https://github.com/tauri-apps/tao/issues/899), [tao#1046](https://github.com/tauri-apps/tao/issues/1046)). The event box takes every pointer event before the buttons can see one.

Most Linux sessions never hit this: `backend_override` (`main.rs`) rewrites a session-default `GDK_BACKEND=wayland` to `x11`, so the app runs through XWayland with compositor-drawn decorations. The dead buttons appear on a **deliberate** native Wayland run (`ANTIBURN_GDK_BACKEND=wayland`) and in a session without an X server — the same audience the popover-placement fix (#581) points at that escape hatch.

## Fix

A new module, `wayland_titlebar`, clears the `above-child` flag right after the shell builds a decorated window (settings, onboarding, main). GTK then delivers clicks to the buttons; title-bar drag outside the buttons keeps working, as in every GTK application.

- The repair guards on `display().backend().is_wayland()`, following the `anchored-window` crate's pattern, so an X11 session with `GTK_CSD=1` cannot misfire it.
- A failed `EventBox` downcast logs at **warn** level — visible in shipped builds — as the tripwire that tao changed its title bar and the module is stale.
- A `#[cfg(not(target_os = "linux"))]` no-op sibling (the `startup_registration` precedent) keeps the three call sites bare, so macOS and Windows builds still type-check them.
- Upstream removed the event box in tao 0.36 ([tao#1218](https://github.com/tauri-apps/tao/pull/1218)); no stable tauri 2.x ships it. The module documents its own removal condition, and that every decorated window must call it.

Also adds the `[Unreleased]/Fixed` CHANGELOG entry (the file is release-workflow-load-bearing) and a native-Wayland title-bar row to the main-window runbook, covering the now-reachable close policy (hide — or quit when the tray icon is off).

## Platform impact

- **macOS / Windows:** no behavior change — `repair` compiles to a no-op; the `gtk` dependency is Linux-target-only.
- **Linux X11 / XWayland (the default):** no-op behind the backend guard.
- **Linux native Wayland:** title-bar buttons work. Presses starting on a button no longer drag the window — the standard GTK trade, noted in the module doc.

## Verification

- Ubuntu GNOME Wayland: instrumented run confirms the repair applies to each decorated window and a close request hides the settings window for warm reopen; a manual title-bar close was confirmed on an earlier build of the same change.
- Eight-angle review: coverage checked against every window builder in the tree (the three decorated windows are exactly the three repaired; all others are `decorations(false)`); focus-stealing ruled out (GTK 3.24 creates titlebar buttons with `can_focus = FALSE`); threading verified safe on and off the main thread.
- `cargo check` / `clippy -D warnings` / `fmt`, `aislop`, prettier: clean. `tsc` + 1930 frontend tests pass. `cargo test`: 1616 pass; the 7 failures in `agent_config`/`insights_report`/`remediation` reproduce identically on `origin/main`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Close and Minimize buttons for the main, Settings, and setup windows when running on native Wayland.
  * Ensured title-bar buttons work in Wayland sessions without an X server and when Wayland is explicitly configured.

* **Tests**
  * Expanded native smoke-test coverage for Wayland title-bar button behavior, including Close behavior aligned with tray settings.


apps/desktop/src-tauri/src/wayland_titlebar.rs
```diff
@@ -0,0 +1,57 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
z0w0: need to remove this license

z0w0 (changes requested):

z0w0 (approved):

apps/desktop/src-tauri/src/wayland_titlebar.rs
martyportier: Removed in c16db750 — the file was the only one in the tree with a license header.

For the record on how it got there: this module was originally written against the pre-rewrite history, where the crate was MPL-2.0 and every source file carried that header, so it was copied from the old `settings.rs`. When the fix was re-applied onto the rewritten (MIT) main, the file came across verbatim and the stale header slipped through. Noted for future ports: strip old headers and re-check conventions against a neighboring file on current main.

PR #586, opened 2026-09-18
feat: broaden agent support and check coverage

z0w0:
## What changed

antiburn can now understand more of the local activity produced by today's coding agents, while staying conservative when a source does not provide enough reliable evidence.

### Better agent support

- Added bounded session support for Amp, Copilot CLI, Cline, and Devin Local.
- Expanded discovery and current-resource inventory coverage for Cursor, Copilot, Cline, Kiro, Amp, Antigravity, and Devin/Windsurf.
- Improved Antigravity, Codex, Kiro, and Windsurf source handling, including companion files, WAL-visible data, Cascade discovery, and source fingerprints.
- Added exact source contracts, fixtures, and characterization coverage for supported formats.
- Kept unsupported or incomplete formats fail-closed instead of presenting misleading findings.

### More trustworthy checks

- Improved partial-evidence handling for model, depth, delegation, cache, and resource checks.
- Tightened clean-result eligibility and freshness detection when source data changes.
- Added Devin freshness fingerprints that include all reader-relevant SQLite state.
- Updated coverage revisions, support claims, and user-facing documentation to match actual behavior.

### Burn Check polish

- Current unused skills, built-in tools, and MCP findings no longer show unrelated sample sessions.
- Long configuration values remain readable on single-change and multi-change review screens.
- Cleaned up cost-chart label rendering and expanded regression coverage around review flows.

## Validation

- Rust tests and Clippy passed.
- Frontend tests passed: 1,787 tests.
- Frontend lint, type-check, and formatting passed.
- `aislop scan --changes` passed with 0 issues.
- `git diff --check` passed.

This PR is intentionally opened as a draft for review of the source contracts and support boundaries.

## Summary by CodeRabbit

* **New Features**
  * Added session analysis for Devin Local, Copilot CLI, Amp, Kiro, Cline, and additional agent data formats.
  * Expanded resource inventory across supported agents, MCP configurations, and nested skill directories.
  * Added Burn Check coverage for Devin Local and additional agent-specific findings.

* **Improvements**
  * Windsurf is now displayed as Devin while retaining its existing integration.
  * Long configuration values in review dialogs wrap within the available width.
  * Improved source validation, fingerprinting, and change detection.

* **Bug Fixes**
  * Empty failed-session views no longer display misleading messages.
  * Improved handling of partial, unreadable, changed, and malformed session data.


daveslutzkin (approved):

codesoda (changes requested):
The broader source support is valuable, but several new routes can currently publish stale, mismatched, or double-counted evidence. In particular, freshness checks do not cover all SQLite/companion inputs, Amp’s claimed-source path has a replacement race, and some source-specific capability and inventory decisions are inconsistent. I recommend addressing these before merging and adding behavioral tests for each contract boundary.

### Blockers / Major

1. **`crates/antiburn-local/src/discovery/source_version.rs` (`devin_content_fingerprint`)** — The fingerprint hashes only root-session rows, while the reader also consumes child message rows, child models, and ACP companions. Changes to those inputs can alter S findings without invalidating cached evidence or failing claimed-source verification. Build one bounded, deterministic Devin snapshot used by both analysis and fingerprinting, include applicable ACP data, and watch all companion roots.

2. **`crates/antiburn-local/src/discovery/source_version.rs` and `apps/desktop/src-tauri/src/analysis.rs` (`bundle_fingerprint`)** — Cline and Copilot reduce their SQLite companion to the main file’s `mtime:size`. Committed WAL-only changes remain visible to readers while leaving that fingerprint unchanged; unrelated checkpoints also invalidate every session. Centralize a session-scoped fingerprint over the exact rows consumed in a read transaction, and test with an open WAL connection and an uncheckpointed commit.

3. **`crates/antiburn-local/src/analysis/vendors/amp.rs` (`visit_claimed`)** — The method validates a `PinnedSource` but calls `self.visit()`, which reopens the path. If the path is replaced after claiming, the replacement is parsed while the original handle can still pass `recheck_full()`. Parse directly from `pinned.reader(...)`, enforce the export-size limit there, and recheck that same consumed handle.

4. **`apps/desktop/src-tauri/src/analysis.rs` (`is_amp_thread_path`)** — Classification searches only forward-slash strings, so native `%APPDATA%\amp\threads\...` paths fall through to `Uncharacterized` and never reach the Amp reader. Match `Path` components or reuse separator normalization, with a native-Windows path test.

5. **`crates/antiburn-local/src/analysis/vendors/devin.rs` (`query_rows`, `acp_companion_index`)** — Devin copies unrestricted SQLite text rows and reads every ACP file fully before filtering. A large row or unrelated companion corpus can exhaust memory and I/O, despite the declared `oversized_row` boundary. Add row/file/record/cumulative-byte limits, stream ACP input, scope lookup to the target session, and report partial coverage when limits are reached.

6. **`apps/desktop/src-tauri/src/agents.rs` (`supports_analysis`)** — Devin is dispatched through `reader_for_input()` by `DevinLocalSqlite`, but capability reporting only checks the agent-level registry, where Windsurf is passive. The backend therefore returns `supports_analysis: false`, overriding the frontend’s `true`. Make capability reporting source-aware, or explicitly report Windsurf/Devin as supported while legacy Windsurf formats remain fail-closed.

7. **`crates/antiburn-local/src/analysis/vendors/amp.rs` (`parse_export`)** — Assistant usage records are emitted without validating a stable message identity, so duplicate records double-count tokens and costs even though `duplicate_usage` is declared invalid. Validate and deduplicate the v39 message identity, reject duplicates, and add an executable duplicate fixture.

8. **`crates/antiburn-local/src/analysis/vendors/copilot.rs` (`visit_bundle` / `visit_reader`)** — Bundled sources set `directory_id` to `None`, bypassing the check between `session.start.data.sessionId` and the `<uuid>` directory. This permits an event stream for one session to be combined with database rows for another. Derive the directory ID from `events_path` and require the directory, event, and `input.session_id` identities to agree.

9. **`apps/desktop/src-tauri/src/agent_config/inventory.rs` (`add_json_mcp_document`, `copilot_inventory`)** — `.github/mcp.json` uses a top-level `servers` map, but the generic parser recognizes only `mcpServers` and `amp.mcpServers`. A normal Copilot file is interpreted as fake servers named `servers`/`inputs`, producing incorrect inventory findings. Parse this file’s actual schema explicitly and test a representative `{"servers": ...}` document.

### Minor

- **`crates/antiburn-local/tests/source_contracts.rs`** — The contract fixtures mostly compare case names or verify schemas they just created; they do not run production readers. The Copilot “valid” schema even omits required `parent_tool_call_id` and `model` columns. Drive these cases through production validators/readers.
- **`crates/antiburn-local/src/analysis/vendors/copilot.rs` (`CopilotState`, `UsageTotals`)** — Event/call identity sets have no total limits, and token aggregation uses unchecked `+=`. Add state caps with partial coverage and use checked aggregation.
- **`crates/antiburn-local/src/analysis/vendors/cline.rs` (`visit_bundle`)** — Each session scans all subagent rows and filters in Rust; fingerprinting repeats the scan. Query by the current parent/session paths and share the session-specific snapshot.
- **`crates/antiburn-local/src/discovery/agents/windsurf.rs` (`discover_devin_sessions`)** — Move hidden/recent-session filtering into SQL to avoid reading the full historical session table on every discovery pass.

### Nits / Docs

- **`docs/check-coverage.md`** — Pi marks C unsupported in the main matrix but later says C remediation is available. Align the matrices and prompt table.
- **`docs/support.md` / `docs/session-coverage.md`** — Copilot is documented as S/O-only and lacking request-depth evidence, while runtime eligibility and the check matrix enable D. Choose one supported contract and document it consistently.
- **`docs/check-coverage.md`** — Replace “Phase 8 inventory files” and the matching test name with baseline behavior terminology.

### Verdict

**Request changes** — the current freshness, source-identity, and platform-classification gaps can produce stale or incorrect findings on newly supported sources.

